PR #28, opened 2024-03-05
Add "Langchain integration" to Future Roadmap

webermatias:
I thought it was a good idea to add to the roadmap a Langchain integration to use Skyvern as a tool that will allow the opensource community to use the browser agent.


suchintan:
This is a brilliant idea

PR #30, opened 2024-03-05
Update run ui script to allow presets + update streamlit to accept them

suchintan:


run_ui.sh
```diff
@@ -1,3 +1,5 @@
+
+
 source "$(poetry env info --path)/bin/activate"
 python scripts/tracking.py skyvern-oss-run-ui
```
wintonzheng, line 4: Why not move the tracking into `streamlit_app/visualizer/streamlit.py`?

suchintan: This PR has nothing to do with tracking

PR #89, opened 2024-03-16
Update README.md

eltociear:
a -> an


ykeremy:
Just noticed this 😅 Thanks for taking the time to open a PR, really appreciate it ❤️ 

When I read the complete sentence, this change feels good as well, WDYT?

> with anti-bot detection mechanisms, proxy networks, and CAPTCHA solving to allow you to complete more complicated workflows.

Issue #68, opened 2024-03-14
Add quickstart instructions for Linux and Windows platforms


ykeremy:
The current quickstart instructions and the setup.sh script is written for and tested on MacOS.

There is nothing that prevents Skyvern from working successfully on other platforms.

It'd be great if we could add instructions for other platforms.

PR #96, opened 2024-03-17
Pause shell when setup fails

nickm980:
Allow the user to read error messages when setup fails for missing dependencies


ykeremy:
Oh yes! I had the same issue the other day while debugging setup script, really appreciate the PR.

I think the main thing is that we don't exit the setup script if something fails. For example, we run `poetry install` but we don't check if it succeeded or failed.
Code references:
- [the code exits if a required command doesn't exist](https://github.com/Skyvern-AI/skyvern/blob/30c5cb9649b5beec135592d99fc1e097894b2844/setup.sh#L19)
- [it ignores all the errors while running poetry install](https://github.com/Skyvern-AI/skyvern/blob/30c5cb9649b5beec135592d99fc1e097894b2844/setup.sh#L161)

We need to do 2 things here:
- make sure we handle failures within each function. An example would look like:
```diff
# Function to install dependencies
install_dependencies() {
-    poetry install
+    if ! poetry install; then
+        echo "Failed to install dependencies."
+        return 1
+    fi
}
```
- update the main function so that we exit the setup script and log the issue if a function returns with a non-zero code

ykeremy:
@nickm980 I just updated the setup script. Now it stop executing next steps if there's a blocking error and tells you why it failed

LawyZheng:
> @nickm980 I just updated the setup script. Now it stop executing next steps if there's a blocking error and tells you why it failed

just return will only return to the main function, will not stop the whole setup (if this is what you expect, then ignore it).

ykeremy:
> > @nickm980 I just updated the setup script. Now it stop executing next steps if there's a blocking error and tells you why it failed
> 
> just return will only return to the main function, will not stop the whole setup (if this is what you expect, then ignore it).

I updated the main function as well, if the functions we call within main return a non-zero value, we log that that step failed and return a nonzero value from the main function as well.

Check it out: https://github.com/Skyvern-AI/skyvern/pull/96/files#diff-4209d788ad32c40cbda3c66b3de47eefb929308ca703bb77a6382625986add17R341-R361

nickm980:
https://github.com/Skyvern-AI/skyvern/assets/81270095/55fc58bc-ca0b-44f8-aa79-cea69680e570

On windows I'm still not able to see what went wrong during installation process

LawyZheng:
> https://github.com/Skyvern-AI/skyvern/pull/96/files#diff-4209d788ad32c40cbda3c66b3de47eefb929308ca703bb77a6382625986add17R341-R361
It's good to return main if the called function returned with an error.
But the setup.sh now is including too many functions, so we can't do some part of setup alone. for example, I just want to update my LLV provider alone or something else.
I already mentioned this in issue https://github.com/Skyvern-AI/skyvern/issues/103 and also created a PR https://github.com/Skyvern-AI/skyvern/pull/102

LawyZheng:
> Command.Prompt.2024-03-18.22-35-10.mp4 
> On windows I'm still not able to see what went wrong during installation process, in this situation I didn't have poetry installed

setup doesn't install poetry, you need to install poetry first, and then execute setup.
How to install peotry, for detail: https://python-poetry.org/docs/#installing-with-the-official-installer

ykeremy:
> Command.Prompt.2024-03-18.22-35-10.mp4 
> On windows I'm still not able to see what went wrong during installation process, in this situation I didn't have poetry installed

Bash scripts don't work on Windows directly, sorry about that. You can try running it on WSL or you can try installing Cygwin and try running it through there. We have plans to extend our quickstart to support other platforms as well, feel free to subscribe to this issue: https://github.com/Skyvern-AI/skyvern/issues/68

ykeremy:
> > @nickm980 I just updated the setup script. Now it stop executing next steps if there's a blocking error and tells you why it failed
> 
> just return will only return to the main function, will not stop the whole setup (if this is what you expect, then ignore it).

This version works as expected on my local. For example, I added a random command to `check_if_required_commands_exist` and ran the script, here's the result:
```
> ./setup.sh
Setting up Skyvern...
Checking for required commands exist...
Error: The following commands are required but not found: random-command
> echo $?   
1

```

ykeremy:
> > https://github.com/Skyvern-AI/skyvern/pull/96/files#diff-4209d788ad32c40cbda3c66b3de47eefb929308ca703bb77a6382625986add17R341-R361
> 
> It's good to return main if the called function returned with an error. But the setup.sh now is including too many functions, so we can't do some part of setup alone. for example, I just want to update my LLV provider alone or something else. I already mentioned this in issue #103 and also created a PR #102

We can still group those functions, let's discuss it under PR #102

PR #104, opened 2024-03-18
add db api to query organizations by name

LawyZheng:


wintonzheng:
@LawyZheng 
what do you want to do with this interface? we don't want to expose such an API for sure

LawyZheng:
Maybe fuzzy search from org.name is more suitable.
What if I lost my org.id or the cred token in secrets.toml?
Now, we can only get org by org.id (which is generated value, not readable by human).
So if users lost org.id or secretes for some reasons, then they need to recreate a new org or find it in db.

In my case, I have multi local env. I registered the org by the same name (org.id is different for sure in different env). Every time switch my dev env, I need to search org.id in the db if I hardcode it(Similar to the situation where I lost my org.id).

wintonzheng:
if it's a problem from the setup script, should fix setup script

LawyZheng:
I think it's not a problem. It offers a more flexible way for some use cases. 
I don't understand what you mean by "fix setup script" --- something like to write a script such as `psql <SQL to find organization by name>` ??

wintonzheng:
@LawyZheng would love to see how this interface is being used within the agent or streamlit/UI or the setup script. Without a good use case, I would prefer not introducing unused code!

PR #112, opened 2024-03-19
Initialize frontend app

msalihaltun:
- Create app with `vite create`
 - Configure ESLint, Typescript
 - Configure husky for pre-commit actions
 - Configure github workflow for frontend

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,7 +100,7 @@ jobs:
         working-directory: ./skyvern-frontend
     steps:
       - name: Check out Git repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Set up Node.js
         uses: actions/setup-node@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,3 +92,26 @@ jobs:
 
       # Finally, run pre-commit.
       - uses: pre-commit/action@v3.0.0
+  
+  fe-lint-build:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./skyvern-frontend
+    steps:
+      - name: Check out Git repository
+        uses: actions/checkout@v3
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: Install Node.js dependencies
+        run: npm ci
+
+      - name: Run linter
+        run: npm run lint
+      
+      - name: Run build
+        run: npm run build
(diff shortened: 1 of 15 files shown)


skyvern-frontend/.husky/pre-commit
```diff
@@ -0,0 +1,3 @@
+cd skyvern-frontend
+
+npm run precommit
```
suchintan, line 3: will this cause issues if we already have pre-commit running?

msalihaltun: This will run the script named `precommit` inside `package.json` in `skyvern-frontend` directory, so I don't think there is a conflict.

ykeremy: hm, can (should?) we include this in our current pre-commit config?

Sth like this within `.pre-commit-config.yaml`
```
.
.
.
  - repo: local
    hooks:
      - id: alembic-check
        name: Alembic Check
        entry: ./run_alembic_check.sh
        language: script
        stages: [ commit ]
  - repo: local
    hooks:
      - id: frontend-precommit
        name: Frontend Pre-commit Hook
        entry: sh -c 'cd skyvern-frontend && npm run precommit'
        language: system
        files: '\.(js|jsx|ts|tsx)$'
        stages: [commit]
```

msalihaltun: The script will have to include `npm install` as well to install `eslint` and `prettier`. Husky precommit hook is lazily set up inside the frontend directory once `npm install` is run.

.github/workflows/ci.yml
```diff
@@ -92,3 +92,26 @@ jobs:
...
       - uses: pre-commit/action@v3.0.0
+  
+  fe-lint-build:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./skyvern-frontend
+    steps:
+      - name: Check out Git repository
+        uses: actions/checkout@v3
```
ykeremy, line 103:
```suggestion
        uses: actions/checkout@v4
```

skyvern-frontend/.eslintrc
```diff
@@ -0,0 +1,27 @@
+{
+  "root": true,
+  "env": { "browser": true, "es2020": true },
+  "extends": [
+    "eslint:recommended",
+    "plugin:@typescript-eslint/recommended",
+    "plugin:react-hooks/recommended",
+    "prettier"
```
suchintan, line 8: Should we also use prettier -recommended?

https://www.reddit.com/r/node/comments/14jh6hz/do_you_use_eslintconfigairbnb_if_so_why_if_not_why/

msalihaltun: Yes, looks like that's the recommended way.

PR #115, opened 2024-03-20
Component library setup

msalihaltun:
- Tailwind CSS
  - shadcn/ui
  - Theme, color scheme

Most of this was generated by [this getting started guide](https://ui.shadcn.com/docs/installation/vite).


msalihaltun:
Had to delete and create `package-lock.json` from scratch to handle [this](https://github.com/vitejs/vite/issues/15167) so the diff is a bit messed up, the actual work should have only been additive 😅

skyvern-frontend/src/App.tsx
```diff
@@ -1,5 +1,12 @@
+import { Button } from "./components/ui/Button";
+import { ThemeProvider } from "@/components/ThemeProvider";
+
 function App() {
-  return <></>;
+  return (
+    <ThemeProvider defaultTheme="dark" storageKey="skyvern-theme">
+      <Button variant={"destructive"}>Hello Shadcn button!</Button>
```
wintonzheng, line 7: 🎉 first you app!

maybe 😃 
```suggestion
      <Button variant={"destructive"}>Hello Skyvern!</Button>
```

skyvern-frontend/src/components/ThemeProvider.tsx
```diff
@@ -0,0 +1,51 @@
+import { useEffect, useState } from "react";
+import { Theme, ThemeProviderContext } from "./themeProviderContext";
+
+export type ThemeProviderProps = {
+  children: React.ReactNode;
+  defaultTheme?: Theme;
+  storageKey?: string;
+};
+
+export function ThemeProvider({
```
wintonzheng, line 10: is this ThemeProvider generated?
I would love to see the UI screenshot if any!

msalihaltun: It's from [here](https://ui.shadcn.com/docs/dark-mode/vite) but I didn't add the toggle for it, I just wanted to have dark by default like the streamlit app currently has.

skyvern-frontend/src/App.tsx
msalihaltun: Great, added.

wintonzheng (approved):

skyvern-frontend/src/util/utils.ts
```diff
@@ -0,0 +1,6 @@
+import { type ClassValue, clsx } from "clsx";
+import { twMerge } from "tailwind-merge";
+
+export function cn(...inputs: ClassValue[]) {
```
wintonzheng, line 4: nit: what does `cn` stand for? can we get a better function name?

msalihaltun: Stands for `classnames`. It's a simple string utility to easily merge css classnames, include / exclude them conditionally, deduplicate etc.

I think it's a convention of this particular library, but there is no harm in renaming it, and I agree it can be confusing. I'll change the name.

msalihaltun: On second thought, maybe I won't break the convention. Because when I use the CLI to add components, it will assume it uses the same `cn` name. It seems like people who use this library are used to that, and I wouldn't want to confuse any future collaborators. Maybe I'll come up with another way to work around this later.

wintonzheng: Got it. i'm okay with it as long as we're aligned with the convention

PR #116, opened 2024-03-21
Add a method for chromium-attached browser_key

illgitthat:
Work in progress, sharing this in case anyone sees what I am missing.


skyvern/webeye/browser_factory.py
```diff
@@ -89,7 +89,13 @@ async def create_browser_context(
             creator = cls._creators.get(browser_type)
             if not creator:
                 raise UnknownBrowserType(browser_type)
-            return await creator(playwright, **kwargs)
+            if browser_type == "chromium-attached":
```
suchintan, line 92: I don't think you need this if statement -- I think you can move the `ws_endpoint` code into `_create_Attached_chromium`

skyvern/webeye/browser_factory.py
```diff
@@ -122,8 +128,27 @@ async def _create_headful_chromium(playwright: Playwright, **kwargs: dict) -> tu
...
+    )
+    contexts = browser.contexts
+    if contexts:
+        browser_context = contexts[0]
+    else:
+        print("No existing contexts found. Creating a new context...")
+        browser_context = await browser.new_context(**browser_args)
+
+    browser_artifacts = BrowserContextFactory.build_browser_artifacts(har_path=browser_args["record_har_path"])
+    return browser_context, browser_artifacts
```
suchintan, line 147: What's the error you get trying this out??

suchintan: This code generally looks correct to me (minus the `if contexts` part)

skyvern/webeye/browser_factory.py
illgitthat: In hindsight, that seems obvious 👍

skyvern/webeye/browser_factory.py
illgitthat: Resharing here:

```2024-03-21T03:26:11.616961Z [info     ] Creating browser state for task task_id=tsk_237547645102207568
2024-03-21T03:26:13.376947Z [info     ] Creating a new page
2024-03-21T03:26:13.917403Z [info     ] A new page is created
2024-03-21T03:26:13.917509Z [info     ] Navigating page to https://www.finditparts.com and waiting for 3 seconds
2024-03-21T03:26:17.559109Z [info     ] Successfully went to https://www.finditparts.com```

Essentially it opens the page and then just stops there.

Is there any log setting to modify to get more verbose logging?

suchintan: Are you connecting over CDP locally? If you turn headless=true does a new window actually open up?

suchintan: Debugging playwright has definitely been a nightmare for us

illgitthat: Yes, connecting locally.

1. Kill all running chromium processes
2. Launch a new window by adding the `--remote-debugging-port=9222`

That pointed me in the right direction. It looks like a lot of the settings needed (like har path) can't be set on an existing browser_context: https://playwright.dev/python/docs/api/class-browsercontext

Although, if I remove the check for using an existing browser context it doesn't launch at all.

suchintan:
@illgitthat This is a great idea -- I'd love to get this merged if you're still open to pushing it across the finish line!

illgitthat:
@suchintan last time I looked at this I believe it isn't possible due to the limitations in the playwright BrowserContext class needed for an existing object.

suchintan:
I'd love to get this merged @illgitthat

illgitthat:
> I'd love to get this merged @illgitthat if you're open to rebasing it to the latest. This is a fantastic idea

I believe it's not currently feasible due to:

> It looks like a lot of the settings needed (like har path) can't be set on an existing browser_context: https://playwright.dev/python/docs/api/class-browsercontext

I will go ahead and close out this PR

PR #119, opened 2024-03-22
Update setup.sh to make sure the poetry / python checks are referenced in the main

suchintan:


setup.sh
```diff
@@ -12,13 +12,15 @@ command_exists() {
     command -v "$1" &> /dev/null
 }
 
-# Ensure required commands are available
-for cmd in poetry python3.11; do
```
suchintan, line 16: removed python 3.11 here because that's installed via poetry later

PR #120, opened 2024-03-23
Tasks page implementation

msalihaltun:
- Implement app layout
 - List tasks in a table
 - Implement task form and samples
 - Implement create task


skyvern-frontend/src/routes/root/Sidebar.tsx
```diff
@@ -1,9 +1,66 @@
...
+  ListBulletIcon,
+  PlusCircledIcon,
+} from "@radix-ui/react-icons";
+import { Link, useLocation } from "react-router-dom";
 
 function Sidebar() {
+  const location = useLocation();
+  let page = "tasks";
+
+  if (location.pathname.includes("create")) {
```
ykeremy, line 13: should we do exact matching instead?

msalihaltun: Locations can be dynamic, eventually will include `task_id` and `step_id`, so exact matching is not possible.

But this is not the right way to control the active state of a navigation component, I'll update it to use `NavLink`.

Issue #147, opened 2024-04-02
Docker Needs to be running 


johnisanerd:
Flagging that docker needs to be running, and that's not covered in the setup instructions.

ykeremy:
Thank you for flagging this! 

Turns out [we just added a log to the setup script](https://github.com/Skyvern-AI/skyvern/blob/999eda9b5d750af6989ccbc9b4ef0a73f7ee8582/setup.sh#L201) 😅 

Would you mind adding a sentence to the quickstart for that one? If not we'll take care of it in the next couple of days.

suchintan:
This is my bad @johnisanerd -- we'd love a quickstart contribution here if you're up for it :)

martincarapia:
@johnisanerd what command did you use to start docker again? Or did you use docker desktop

suchintan:
@martincarapia definitely use docker desktop!

martincarapia:
Ok, because I see online there are CLI ways to run Docker Daemon. Although I believe Desktop is more suitable since it covers Windows, macOS, and Linux

PR #153, opened 2024-04-04
Update README.md to include Docker. 

martincarapia:
Just fill out the docker start description with the container_name/container_id, and then it should be ready for prod. 
"
Run docker
   ```bash
   docker start [options] [container_name/container_id]
   ```
"
Closes issue #147


README.md
```diff
@@ -91,6 +92,10 @@ Note: Our setup script does these two for you, but they are here for reference.
     ```bash
     ./setup.sh
     ```
+1. Run docker
+   ```bash
+   docker start [options] [container_name/container_id]
+   ```
```
ykeremy, line 98: This is the command to actually start running a docker container.

I thought the issue was the Docker Daemon / Docker Desktop not running. 

We need an action item here that would prevent this check from failing:
https://github.com/Skyvern-AI/skyvern/blob/999eda9b5d750af6989ccbc9b4ef0a73f7ee8582/setup.sh#L199-L203

ykeremy (commented):
👋🏼 thank you so much for the PR

README.md
ykeremy: Yes, exactly! 

If you could tell how people can run it for each OS that would be perfect. I know it's not always easily possible from the CLI. So for windows you could just say run Docker Desktop, etc.

README.md
```diff
@@ -91,6 +92,7 @@ Note: Our setup script does these two for you, but they are here for reference.
     ```bash
     ./setup.sh
     ```
+1. Open Docker Desktop (Works for Windows, macOS, and Linux) or run Docker Daemon
```
ykeremy, line 95: Thanks for the update, looks good.

Can you also move this line before the setup.sh?

So it should be:
1. clone the repo
2. run docker desktop
3. setup.sh 
4. ...
5. ...
6. ...

ykeremy (approved):

ykeremy:
Thanks for the PR @martincarapia, I'm gonna merge it.

PR #162, opened 2024-04-07
docker compose setup

LawyZheng:
- build a docker image
- support docker compose setup

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,6 +137,9 @@ jobs:
       uses: aws-actions/amazon-ecr-login@v2
       with:
         registry-type: public
+    
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v1
 
     - name: Build, tag, and push image to Amazon Public ECR
       id: build-image
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,10 @@ on:
   push:
     branches: [main]
 
+env:
+  AWS_REGION: us-east-1
+  ECR_REPOSITORY: skyvern
+
 jobs:
   test:
     runs-on: ubuntu-latest
@@ -115,3 +119,33 @@ jobs:
       
       - name: Run build
         run: npm run build
+
+  build-docker-image:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ env.AWS_REGION }}
+
+      - name: Login to Amazon ECR
+        id: login-ecr
+        uses: aws-actions/amazon-ecr-login@v1
+
+      - name: Build, tag, and push image to Amazon Public ECR
+        id: build-image
+        uses: docker/build-push-action@v2
+        env:
+          ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
+          IMAGE_TAG: ${{ github.sha }}
+        with:
+          context: .
+          platforms: |
+            linux/amd64
+            linux/arm64
+          push: true
+          tag: |
+            ${{ env.ECR_REGISTRY}}/${{ env.ECR_REPOSITORY }}:${{ env.IMAGE_TAG }}
+            ${{ env.ECR_REGISTRY}}/${{ env.ECR_REPOSITORY }}:latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,14 +4,10 @@ name: Run tests and pre-commit
 # `branches: [main], then this actions runs _twice_ on pull requests, which is
 # annoying.
 on:
+  workflow_call:
   pull_request:
   push:
     branches: [main]
-
-# env:
-#   AWS_REGION: us-east-1
-#   ECR_REPOSITORY: skyvern
-#   REGISTRY_ALIAS: t6d4b5t4 # skyvern
 
 jobs:
   test:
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -115,3 +119,34 @@ jobs:
       
       - name: Run build
         run: npm run build
+
+  build-docker-image:
```
wintonzheng, line 123: we only want to run `build-docker-image` when a new commit is merged to `main`.

we shouldn't run on `pull_request` (line 7)

maybe this job should be in a separate yml file in order to do that?

LawyZheng: actually, I think this ci should be triggered when we release a version.
anyway, I just use this for testing building and pushing.
I will fix this problem after everything is going well

wintonzheng: ah you're right

.github/workflows/ci.yml
```diff
@@ -115,3 +119,34 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: [test, fe-lint-build]
+    steps:
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@v1
```
wintonzheng, line 128: update to v2?
see https://github.com/aws-actions/amazon-ecr-login?tab=readme-ov-file#new-v2-release

docker-compose.yml
```diff
@@ -0,0 +1,57 @@
...
+      - ./artifacts:/data/artifacts
+      - ./videos:/data/videos
+      - ./har:/data/har
+      - ./.streamlit:/app/.streamlit
+    environment:
+      - DATABASE_STRING=postgresql+psycopg://skyvern:skyvern@postgres:5432/skyvern
+      - BROWSER_TYPE=chromium-headful
+      - ENABLE_OPENAI=true
+      - OPENAI_API_KEY=<your_openai_key>
+      # If you want to use other LLM provide, like azure or anthropic
```
wintonzheng, line 33:
```suggestion
      # If you want to use other LLM provider, like azure and anthropic:
```

docker-compose.yml
```diff
@@ -0,0 +1,57 @@
...
+    restart: on-failure
+    # comment out if you want to externally call skyvern API
+    # ports:
+    #   - 8000:8000
+    volumes:
+      - ./artifacts:/data/artifacts
+      - ./videos:/data/videos
+      - ./har:/data/har
+      - ./.streamlit:/app/.streamlit
+    environment:
```
wintonzheng, line 28: can we point it to the `.env` file?

wintonzheng (commented):
LGTM overall. Thanks for building this!

docker-compose.yml
LawyZheng: If `.env` pointed, users need to run the docker-compose together with `.env`.
In this way, rookie users could run the single `docker-compose.yml` file.
And advanced users could point it to the `.env` by themselves.

LawyZheng: Besides, if we change the settings in env, we are supposed to restart/recreate the container.
If we have an isolated env setting file, I think it's easy to forget to restart/recreate.
I think load from `.env` could be more useful if we want to share the same env between multiple containers, rather than running a single container

docker-compose.yml
```diff
@@ -0,0 +1,59 @@
...
+      # - ENABLE_ANTHROPIC=true
+      # - ANTHROPIC_API_KEY=<your_anthropic_key>
+      # - ENABLE_AZURE=true
+      # - LLM_KEY=AZURE_OPENAI_GPT4V
+      # - AZURE_DEPLOYMENT=<your_azure_deployment>
+      # - AZURE_API_KEY=<your_azure_api_key>
+      # - AZURE_API_BASE=<your_azure_api_base>
+      # - AZURE_API_VERSION=<your_azure_api_version>
+    depends_on:
+      - postgres
```
ykeremy, line 43: This wouldn't work well because this way it just waits for the container to be up. Instead, we should be waiting for postgres to be ready. 

This eventually works out but until skyvern is up, the container will restart multiple times since it'll die without a healthy DB.

README.md
```diff
@@ -101,6 +101,16 @@ Note: Our setup script does these two for you, but they are here for reference.
     ```
 1. Navigate to `http://localhost:8501` in your browser to start using the UI
 
+## Docker Compose setup
+
+1. Fill in the LLM provider key on the [docker-compose.yml](./docker-compose.yml)
+2. Run the following command:
+   ```bash
+    docker compsoe up -d
```
ykeremy, line 109:
```suggestion
    docker compose up -d
```

README.md
```diff
@@ -101,6 +101,16 @@ Note: Our setup script does these two for you, but they are here for reference.
     ```
 1. Navigate to `http://localhost:8501` in your browser to start using the UI
 
+## Docker Compose setup
```
ykeremy, line 104: We should test the docker, (maybe even ask people in DC to test it out and provide feedback) and when it's all good this should be the primary way to start using Skyvern.

docker-compose.yml
LawyZheng: Great, I have noticed it already.
But I don't have any idea how to make skyvern check the DB health hook in docker-compose.
Any good idea?

LawyZheng: So I just let skyvern restart on failure. haha

ykeremy: Yeah it works but if someone checks out the logs or runs without `-d` it could get confusing 😄 

Can you check this out and see if it works? I did something similar before, if this doesn't work we can hop on a call later?

https://stackoverflow.com/a/55835081

LawyZheng: It worked.

ykeremy:
@LawyZheng the Skyvern container waits the postgres to actually be ready now!! 🎉 

Stearmlit can't connect to the server though:
``` 
ConnectionError: HTTPConnectionPool(host='127.0.0.1', port=8000): Max retries exceeded with url: /api/v1/internal/tasks?page=1&page_size=15 (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0xffff5e1d8ad0>: Failed to establish a new connection: [Errno 111] Connection refused'))
```
  
LMK if you need help debugging.

LawyZheng:
@ykeremy  you'd better delete or move away the old `secrets.toml` in folder `.streamlit`.
If you are trying to use the old DB, you need to verify some things:
1. the connect string in docker-compose is correct.
2. need change the the host in `secrets.toml` to `skyvern` from `127.0.0.1`/ or you can map a port 8000 to the local machine in the skyvern container

LawyZheng:
My default docker-compose is prepared for uses who haven't do any setups yet.
And only streamlit port is exposed to the local machine network. postgres and skyvern are both in the docker network without exported.

andr0s:
btw this seems to be working, i've just tested it :) Thanks

PR #164, opened 2024-04-07
Different portsUpdate ./run_skyvern.sh to run both the visualizer and the server on different ports  is completed

unaisshemim:


suchintan:
@unaisshemim are you able to update this PR? There are a few errors inside it

PR #206, opened 2024-04-18
support dropdown on combobox

LawyZheng:
- open the dropdown before scraping so we can get the options in the scraped_page.
- send the option value to LLM, helping it to make the right choice.

### DIFF
--- a/skyvern/webeye/scraper/domUtils.js
+++ b/skyvern/webeye/scraper/domUtils.js
@@ -378,6 +378,23 @@ function isInteractable(element) {
   return false;
 }
 
+const isComboboxDropdown = (element) => {
+  if (element.tagName.toLowerCase() !== "input") {
+    return false;
+  }
+  const role = element.getAttribute("role")
+    ? element.getAttribute("role").toLowerCase()
+    : "";
+  const haspopup = element.getAttribute("aria-haspopup")
+    ? element.getAttribute("aria-haspopup").toLowerCase()
+    : "";
+  const readonly =
+    element.getAttribute("readonly") &&
+    element.getAttribute("readonly").toLowerCase() !== "false";
+  const controls = element.hasAttribute("aria-controls");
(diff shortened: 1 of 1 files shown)


skyvern/webeye/scraper/domUtils.js
```diff
@@ -601,6 +618,21 @@ function buildTreeFromBody() {
     } else if (attrs["role"] && attrs["role"].toLowerCase() === "listbox") {
       // if "role" key is inside attrs, then get all the elements with role "option" and get their text
       selectOptions = getListboxOptions(element);
+    } else if (isComboboxDropdown(element)) {
+      // open combobox dropdown to get options
+      element.click();
```
wintonzheng, line 623: very interesting - interacting with the dropdown elements in order to get/parse dropdown elements.

would element ids / unique ids be assigned to the dropdown elements? would line 629 (pressing tab) makes those elements gone. if that happens, then the dropdown element ids we're sending to llm won't be interactable anymore

wintonzheng: makes sense

skyvern/webeye/scraper/domUtils.js
```diff
@@ -601,6 +618,21 @@ function buildTreeFromBody() {
...
+      const listBox = document.getElementById(
```
wintonzheng, line 624: is it possible that this listbox will be undefined / null? shall we handle that

wintonzheng (approved):
logic looks good. let's make sure we cover edge cases. For example, handle all the null cases in this PR so that it won't break dom parsing

skyvern/webeye/scraper/domUtils.js
LawyZheng: Yeah, you're right. We should handle the null exception here, tho the possibility to be null is very low.

PR #228, opened 2024-04-24
Use first page of running tasks since api is not filtered

msalihaltun:


msalihaltun:
This will need to be fixed when the tasks endpoint has filters for status

PR #230, opened 2024-04-24
GPT-4-Turbo w/ Vision is GA now

suchintan:


ykeremy:
You also need to set supports vision True

suchintan:
You're right

PR #232, opened 2024-04-24
Allow querying multiple statuses

msalihaltun:


ykeremy:
Not sure if we have a DB index on task.status, can you check?

msalihaltun:
> Not sure if we have a DB index on task.status, can you check?

Yes we do

wintonzheng (approved):

wintonzheng (approved):

PR #282, opened 2024-05-08
Fix setup.sh to include the last anthropic models

OB42:
Hi 👋 
the setup.sh file seems outdated for Anthropic given that there are 3 different models now, it was failing with error `skyvern.forge.sdk.api.llm.exceptions.InvalidLLMConfigError: LLM config with key ANTHROPIC_CLAUDE3 is not a valid LLMConfig`
here is a fix 😄


suchintan (approved):

suchintan:
Ah, we introduced a backwards incompatible change here: https://github.com/Skyvern-AI/skyvern/commit/4459ca7d096edfea27fd0be8f321672d898f1e12

Thanks for fixing!

PR #285, opened 2024-05-09
Default the old anthropic model name for backwards compatibility

ykeremy:


suchintan:
@OB42 This will preserve backwards compatibility (https://github.com/Skyvern-AI/skyvern/pull/282)

PR #289, opened 2024-05-09
fix scrolling problem

LawyZheng:


OB42:
IMO using behavior:instant everywhere would probably trigger a lot more captcha/bot detectors.

LawyZheng:
> IMO using behavior:instant everywhere would probably trigger a lot more captcha/bot detectors.

Good idea. but it's more important to take screenshots than worrying about the bot-detection problem.

OB42:
> > IMO using behavior:instant everywhere would probably trigger a lot more captcha/bot detectors.
> 
> Good idea. but it's more important to take screenshots than worrying about the bot-detection problem.

how is scrolling related to screenshots exactly?

wintonzheng:
@OB42 
For a long page, Skyvern does multiple scrolls until the end of the page, to generate multiple screenshots and send those to LLM together. We've tested this heavily earlier which drove us to this solution: LLMs perform better with multiple screenshots compared to sending one big whole page screenshot.

Most of the websites have default auto behavior which jumps instantly when using scrollBy. We haven't seen any bot detection issues so far. Pls let us know if you find a bot detection example and we can probably help figure out a solution.

wintonzheng:
@LawyZheng 
The scrollUpAndDown and scrollDownAndUp can be removed. It was there to counter the weird select-2 behavior which has been fixed now

wintonzheng:
@LawyZheng 
after removing scroll up and down, we might not need the `behavior: instant`. Can you test that?

LawyZheng:
> @LawyZheng after removing scroll up and down, we might not need the `behavior: instant`. Can you test that?

sure. I will test some cases to see if it still works well without `behavior: instant`

wintonzheng (approved):

wintonzheng:
@LawyZheng PR looks good

PR #290, opened 2024-05-09
Force Claude3 models to output JSON object and parse it more reliably

OB42:
Anthropic models seems to struggle to output valid json objects and it was causing Skyvern to fail +/- randomly, sometimes the anthropic output looked like this:

`Here's a JSON object : {...}`

Which caused the function parse_api_response to fail.

We can force it to output json by prefilling the response with "{": [https://docs.anthropic.com/claude/docs/prefill-claudes-response](https://docs.anthropic.com/claude/docs/prefill-claudes-response)

I also changed the way JSON is extracted from markdown, because the way it was done before might be too unreliable if some of the json data contain backticks in a string.

And I removed any text before and after the first and last braces of the data, Anthropic sometimes seemed to struggle to output the extraction format and was outputting text after the object.

Warning: This was only tested using Anthropic(with Claude 3 Opus) as I do not have OpenAI credits at the moment. Feel free to test it/edit it as necessary.

### DIFF
--- a/skyvern/forge/sdk/api/llm/utils.py
+++ b/skyvern/forge/sdk/api/llm/utils.py
@@ -65,4 +65,3 @@ def try_to_extract_json_from_markdown_format(text):
         return match.group(1)
     else:
         return text
-    
--- a/skyvern/forge/sdk/api/llm/utils.py
+++ b/skyvern/forge/sdk/api/llm/utils.py
@@ -52,13 +52,13 @@ def parse_api_response(response: litellm.ModelResponse) -> dict[str, str]:
         raise InvalidLLMResponseFormat(str(response)) from e
 
 
-def replace_useless_text_around_json(input_string):
+def replace_useless_text_around_json(input_string: str) -> str:
     first_occurrence_of_brace = input_string.find("{")
     last_occurrence_of_brace = input_string.rfind("}")
     return input_string[first_occurrence_of_brace : last_occurrence_of_brace + 1]
 
 
-def try_to_extract_json_from_markdown_format(text):
+def try_to_extract_json_from_markdown_format(text: str) -> str:
     pattern = r"```json\s*(.*?)\s*```"
     match = re.search(pattern, text, re.DOTALL)
     if match:
--- a/skyvern/forge/sdk/api/llm/utils.py
+++ b/skyvern/forge/sdk/api/llm/utils.py
@@ -30,16 +30,16 @@ async def llm_messages_builder(
                     },
                 }
             )
-    #Anthropic models seems to struggle to always output a valid json object so we need to prefill the response to force it:
+    # Anthropic models seems to struggle to always output a valid json object so we need to prefill the response to force it:
     if SettingsManager.get_settings().ENABLE_ANTHROPIC:
-        return [{"role": "user", "content": messages},{"role": "assistant", "content": "{"}]
+        return [{"role": "user", "content": messages}, {"role": "assistant", "content": "{"}]
     return [{"role": "user", "content": messages}]
 
 
 def parse_api_response(response: litellm.ModelResponse) -> dict[str, str]:
     try:
         content = response.choices[0].message.content
-        #Since we prefilled Anthropic response with "{" we need to add it back to the response to have a valid json object:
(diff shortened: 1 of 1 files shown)


suchintan:
you're two steps ahead of my discord comment

suchintan:
@OB42 can we try this approach first? https://docs.anthropic.com/claude/docs/control-output-format

OB42:
> @OB42 can we try this approach first? https://docs.anthropic.com/claude/docs/control-output-format
Well I wrote this code after reading that documentation.(the first link I sent is also linked on the documentation you sent, just below that https://docs.anthropic.com/claude/docs/control-output-format#prefilling-claudes-response )

`        return [{"role": "user", "content": messages}, {"role": "assistant", "content": "{"}]` 

The other approach seems to specify the desired format, but that seems to already be the case in the prompts(?)

suchintan:
@OB42 You're right -- that was a silly comment by me!

suchintan:
This is great. @ykeremy can you review?

skyvern/forge/sdk/api/llm/utils.py
```diff
@@ -29,17 +31,37 @@ async def llm_messages_builder(
                     },
                 }
             )
-
+    # Anthropic models seems to struggle to always output a valid json object so we need to prefill the response to force it:
+    if SettingsManager.get_settings().ENABLE_ANTHROPIC:
```
ykeremy, line 35: `ENABLE_ANTHROPIC=True` doesn't mean that we're using Anthropic to send the current LLM request. Also, it's possible to use Anthropic through AWS Bedrock. 

I think a good way to move forward here is to update [LLMConfig](https://github.com/Skyvern-AI/skyvern/blob/240551f68684a42cd54a1af6167bfa5eafe5b049/skyvern/forge/sdk/api/llm/models.py#L9) and implement an `add_assistant_prefix` property. 

This way, we can configure assistant prefill manually and always make sure we are checking the language model provider + language model to make an accurate decision.

Required changes:
- [ ] Add `add_assistant_prefix` field to LLMConfig with a default `False` value
- [ ] Update the existing Anthropic `LLMConfig`s to have this field as `True`
    - Can be found [here](https://github.com/Skyvern-AI/skyvern/blob/240551f68684a42cd54a1af6167bfa5eafe5b049/skyvern/forge/sdk/api/llm/config_registry.py#L61-L100)
- [ ] We're going to need to pass the `add_assistant_prefix` to `llm_message_builder` and check that instead of the `ENABLE_ANTHROPIC/ENABLE_BEDROCK` settings.

ykeremy (changes requested):
Thank you so much for the PR, tested it out on my local as well and it just works perfectly. Well, almost 😅 There is one change we need to make before merging this. 

Let me know if you have questions or you want us to take it over from here in case you don't have time to update the PR.

skyvern/forge/sdk/api/llm/utils.py
suchintan: Maybe we can do this in a follow up? It's possible BEDROCK may have a different schema for defining this, similar to how[ OpenAI's JSON mode](https://platform.openai.com/docs/guides/text-generation/json-mode) is different than [Azure Open AI's ](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/json-mode?tabs=python)

OK It's not really that different, but in theory it could be so we can defer this until later!

ykeremy: Implemented the mentioned changes in #293

ykeremy:
Will merge https://github.com/Skyvern-AI/skyvern/pull/293 with additional changes, closing this PR. @OB42, thank you so much for the contribution

OB42:
> Will merge #293 with additional changes, closing this PR. @OB42, thank you so much for the contribution

You're welcome! :) sorry for not adding the required changes quicker 😅

suchintan:
> > Will merge #293 with additional changes, closing this PR. @OB42, thank you so much for the contribution
> 
> You're welcome! :) sorry for not adding the required changes quicker 😅

This was a fantastic contribution! Keep em coming!

PR #305, opened 2024-05-13
put api key inside curl if available

ykeremy:


msalihaltun:
There is a mistake here, I'll redo this

PR #316, opened 2024-05-14
optimize select action

LawyZheng:
- support select action on <label> of <input>(checkbox/radio)
- use check to avoid double clicks, which will uncheck the checkbox

### DIFF
--- a/skyvern/webeye/actions/handler.py
+++ b/skyvern/webeye/actions/handler.py
@@ -22,6 +22,7 @@ from skyvern.webeye.actions import actions
 from skyvern.webeye.actions.actions import (
     Action,
     ActionType,
+    CheckboxAction,
     ClickAction,
     ScrapeResult,
     SelectOptionAction,
@@ -305,7 +306,7 @@ async def handle_select_option_action(
             click_action = ClickAction(element_id=action.element_id)
             return await chain_click(task, page, click_action, child_anchor_xpath)
 
-        # handler the select action on <laebel>
+        # handler the select action on <label>
         select_element_id = get_select_id_in_label_children(scraped_page, action.element_id)
         if select_element_id is not None:
             LOG.info(
@@ -315,6 +316,17 @@ async def handle_select_option_action(
             )
             select_action = SelectOptionAction(element_id=select_element_id, option=action.option)
             return await handle_select_option_action(select_action, page, scraped_page, task, step)
+
+        # handle the select action on <label> of checkbox/radio
+        checkbox_element_id = get_checkbox_id_in_label_children(scraped_page, action.element_id)
+        if checkbox_element_id is not None:
+            LOG.info(
+                "SelectOptionAction is on <label> of <input> checkbox/radio. take the action on the real <input> checkbox/radio",
+                action=action,
+                checkbox_element_id=checkbox_element_id,
+            )
+            check_action = CheckboxAction(element_id=checkbox_element_id, is_checked=True)
+            return await handle_checkbox_action(check_action, page, scraped_page, task, step)
 
         return [ActionFailure(Exception("No anchor tag or select children found for the label for SelectOptionAction"))]
     elif tag_name == "a":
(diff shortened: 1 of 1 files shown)


skyvern/webeye/actions/handler.py
```diff
@@ -316,6 +317,17 @@ async def handle_select_option_action(
             select_action = SelectOptionAction(element_id=select_element_id, option=action.option)
             return await handle_select_option_action(select_action, page, scraped_page, task, step)
 
+        # handle the select action on <label> of checkbox/radio
+        checkbox_element_id = get_checkbox_id_in_label_children(scraped_page, action.element_id)
```
ykeremy, line 321: before we check children of label, can we check if the which element the label is `for`?

If it's not a checkbox, then we can try looking at its children 🤔

skyvern/webeye/actions/handler.py
```diff
@@ -362,12 +374,13 @@ async def handle_select_option_action(
...
-    ) in ["radio", "checkbox"]:
-        # TODO: double click will uncheck the checkbox
-        click_action = ClickAction(element_id=action.element_id)
-        return await chain_click(task, page, click_action, xpath)
+    elif tag_name == "input" and element_dict.get("attributes", {}).get("type", None) in ["radio", "checkbox"]:
+        LOG.info(
+            "SelectOptionAction is on <input> checkbox/radio",
+            action=action,
+        )
+        check_action = CheckboxAction(element_id=action.element_id, is_checked=True)
```
ykeremy, line 382: With this change we're losing the ability to uncheck checkboxes, right?

skyvern/webeye/actions/handler.py
LawyZheng: how about doing it in another ticket?
since it's a more complicated situation.
Like we might get the label `for`, bonding to the <select>  or <input> in the select action.
both should be handled.

skyvern/webeye/actions/handler.py
LawyZheng: Yes, you're right.
But I think it's ok since we don't expect LLM to return some actions to do the uncheck.
I think we only assume LLM will return actions to check.

ykeremy: For the most usecases today that's correct, it's ok for this PR. Can you create a ticket for it so we won't forget?

skyvern/webeye/actions/handler.py
```diff
@@ -316,6 +317,17 @@ async def handle_select_option_action(
...
 
+        # handle the select action on <label> of checkbox/radio
+        checkbox_element_id = get_checkbox_id_in_label_children(scraped_page, action.element_id)
+        if checkbox_element_id is not None:
+            LOG.info(
+                "SelectOptionAction is on <label> of <input> checkbox/radio. take the action on the real <input> checkbox/radio",
+                action=action,
+                checkbox_element_id=checkbox_element_id,
+            )
+            check_action = CheckboxAction(element_id=checkbox_element_id, is_checked=True)
```
ykeremy, line 328: Does this work with `radio`s?

LawyZheng: yes, check also works on radio button.

ykeremy (approved):

skyvern/webeye/actions/handler.py
```diff
@@ -643,6 +662,22 @@ def get_select_id_in_label_children(scraped_page: ScrapedPage, element_id: int)
...
+    """
+    search checkbox/radio in the children of <label>
+    """
+    LOG.info("Searching checkbox/radio in the label children", element_id=element_id)
+    element = scraped_page.id_to_element_dict.get(element_id, None)
+    if element is None:
+        return None
+
+    for child in element.get("children", []):
+        if child.get("tagName", "") == "input" and child.get("attributes", {}).get("type") in ["checkbox", "radio"]:
```
ykeremy, line 675: is it possible there are multiple? if so I think we should make a decision but log an error.

LawyZheng: According to the HTML rule, there shouldn't be multiple `<input>` or `<select>` in the child of label.
Otherwise `<label>` will lose its default binding.

skyvern/webeye/actions/handler.py
LawyZheng: some elements bond with the `<label>` might be invisible on the page. 
go to https://github.com/Skyvern-AI/skyvern/pull/273 for more details.
and the select and check should be handled separately, so it's a little complicated.

PR #353, opened 2024-05-22
save the elements tree in the prompt as an artifact

LawyZheng:
- our element tree has two format: HTML/JSON
- save current using element format as an artifact. If the JSON format is used, we save the JSON; if the HTML format is used, we save the HTML.

### DIFF
--- a/skyvern/forge/agent.py
+++ b/skyvern/forge/agent.py
@@ -791,6 +791,11 @@ class ForgeAgent:
             artifact_type=ArtifactType.VISIBLE_ELEMENTS_TREE_TRIMMED,
             data=json.dumps(scraped_page.element_tree_trimmed, indent=2).encode(),
         )
+        await app.ARTIFACT_MANAGER.create_artifact(
+            step=step,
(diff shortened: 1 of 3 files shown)


skyvern/forge/agent.py
```diff
@@ -791,6 +791,11 @@ async def _build_and_record_step_prompt(
             artifact_type=ArtifactType.VISIBLE_ELEMENTS_TREE_TRIMMED,
             data=json.dumps(scraped_page.element_tree_trimmed, indent=2).encode(),
         )
+        await app.ARTIFACT_MANAGER.create_artifact(
+            step=step,
+            artifact_type=ArtifactType.VISIBLE_ELEMENTS_TREE_IN_PROMPT,
+            data=scraped_page.build_element_tree(element_tree_format).encode(),
```
wintonzheng, line 797: this and line 772 is duplicated. we can set a variable `element_tree_in_prompt` before line 772 and reuse that variable

wintonzheng (approved):

PR #361, opened 2024-05-24
Use 4-char element ids instead of sequential integers

alexzarbn:
## Problem
This issue with element ids is tightly related to the iframes support. When element tree is built by injecting `domUtils.js` into each iframe, the resulting ids are being messed up, because at first the page element tree is built and then element tree of each iframe on that page, resulting in iframe children ids duplicating some ids of elements on the page.

## Solution
Use random 4-char ids for elements. This way the we can manipulate the tree without worrying about uniqueness of the element ids.


skyvern/webeye/scraper/domUtils.js
```diff
@@ -620,7 +630,7 @@ function buildTreeFromBody() {
   };
 
   function buildElementObject(element, interactable) {
-    var element_id = elements.length;
+    var element_id = element.getAttribute('unique_id') ?? uniqueId();
```
suchintan, line 633: nice I love this change

skyvern/webeye/scraper/domUtils.js
```diff
@@ -563,6 +563,16 @@ function getListboxOptions(element) {
   return selectOptions;
 }
 
+function uniqueId() {
+    const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
```
ykeremy, line 567: I wonder if making the unique IDs case insensitive would prevent LLMs from hallucinating.

ykeremy: I tested with Claude 3 Sonnet and GPT4O on my local, didn't hallucinate so far. Let's keep it this way but we should keep an eye out for it @Skyvern-AI/eng

ykeremy (approved):
Looks great! Tested on my local as well, worked really well. I've couple of small comments.

skyvern/webeye/scraper/domUtils.js
```diff
@@ -767,13 +777,13 @@ function buildTreeFromBody() {
           if (parentId === null) {
             resultArray.push(elementObj);
           } else {
-            elements[parentId].children.push(elementObj);
+            elements.find((element) => element.id === parentId).children.push(elementObj)
```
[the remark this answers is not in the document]
ykeremy: We can write a TODO here and do it later as well, not a huge difference given the # elements. Fixing or adding a todo works.

ykeremy:
There are some prettier and mypy errors as well, would be perfect if you could take a look at those.

skyvern/webeye/scraper/domUtils.js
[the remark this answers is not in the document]
alexzarbn: Added a TODO. Given the number of elements, performance impact is minimal.

alexzarbn:
> There are some prettier and mypy errors as well, would be perfect if you could take a look at those.

Fixed those, but for this `mypy` error I had to [add](https://github.com/Skyvern-AI/skyvern/pull/361/commits/83b4258ca4dd531f85cc3d942b016cf1bab1b996) unique ids to options too. Not sure, if this is the right course of action in this case. 

```bash
skyvern/webeye/actions/handler.py:487: error: Invalid index type "int" for "dict[str, str]"; expected type "str"  [index]
```

Please met know, what was the intention behind mapping index -> id in the prompt, whether it was a typo (also considering that there is no `optionIndex` included in the prompt as far as I can see) or it was intentional, and whether including options ids in the prompt is okay. Perhaps the implementation should rely solely on option indexes and match by xpath could be removed?

```
"index": int, // the id corresponding to the optionIndex under the the select element.
```

skyvern/forge/prompts/skyvern/extract-action.j2
```diff
@@ -13,14 +13,15 @@ Reply in JSON format with the following keys:
...
         "text": str, // Text for INPUT_TEXT action only
         "file_url": str, // The url of the file to upload if applicable. This field must be present for UPLOAD_FILE but can also be present for CLICK only if the click is to upload the file. It should be null otherwise.
         "download": bool, // Can only be true for CLICK actions. If true, the browser will trigger a download by clicking the element. If false, the browser will click the element without triggering a download.
         "option": {  // The option to select for SELECT_OPTION action only. null if not SELECT_OPTION action
             "label": str, // the label of the option if any. MAKE SURE YOU USE THIS LABEL TO SELECT THE OPTION. DO NOT PUT ANYTHING OTHER THAN A VALID OPTION LABEL HERE
-            "index": int, // the id corresponding to the optionIndex under the the select element.
+            "index": int, // the index corresponding to the index under the select element.
             "value": str // the value of the option. MAKE SURE YOU USE THIS VALUE TO SELECT THE OPTION. DO NOT PUT ANYTHING OTHER THAN A VALID OPTION VALUE HERE
+            "id": str // the id of the option.
```
LawyZheng, line 24: Can we remove the option.id to prevent LLM hallucinating.

skyvern/webeye/actions/handler.py
```diff
@@ -484,7 +484,7 @@ async def handle_select_option_action(
             return [ActionFailure(e)]
 
     try:
-        option_xpath = scraped_page.id_to_xpath_dict[action.option.index]
+        option_xpath = scraped_page.id_to_xpath_dict[action.option.id]
```
LawyZheng, line 487: Actually this is a legacy bug. We didn't store the option element as a single key in the `scraped_page.id_to_xpath_dict`.  So the `option.id` here will never be found.

https://github.com/Skyvern-AI/skyvern/blob/418473d72c0d867a97ac543cb7988a45f2e2f275/skyvern/webeye/scraper/scraper.py#L257-L263

LawyZheng:
this can trigger an error. Can you remove this? 
https://github.com/Skyvern-AI/skyvern/blob/418473d72c0d867a97ac543cb7988a45f2e2f275/skyvern/webeye/actions/handler.py#L776

LawyZheng:
> @ykeremy
> 
> > There are some prettier and mypy errors as well, would be perfect if you could take a look at those.
> 
> Fixed those, but for this `mypy` error I had to [add](https://github.com/Skyvern-AI/skyvern/pull/361/commits/83b4258ca4dd531f85cc3d942b016cf1bab1b996) unique ids to options too. Not sure, if this is the right course of action in this case.
> 
> ```shell
> skyvern/webeye/actions/handler.py:487: error: Invalid index type "int" for "dict[str, str]"; expected type "str"  [index]
> ```
> 
> Please met know, what was the intention behind mapping index -> id in the prompt, whether it was a typo (also considering that there is no `optionIndex` included in the prompt as far as I can see) or it was intentional, and whether including options ids in the prompt is okay. Perhaps the implementation should rely solely on option indexes and match by xpath could be removed?
> 
> ```
> "index": int, // the id corresponding to the optionIndex under the the select element.
> ```



> @ykeremy
> 
> > There are some prettier and mypy errors as well, would be perfect if you could take a look at those.
> 
> Fixed those, but for this `mypy` error I had to [add](https://github.com/Skyvern-AI/skyvern/pull/361/commits/83b4258ca4dd531f85cc3d942b016cf1bab1b996) unique ids to options too. Not sure, if this is the right course of action in this case.
> 
> ```shell
> skyvern/webeye/actions/handler.py:487: error: Invalid index type "int" for "dict[str, str]"; expected type "str"  [index]
> ```
> 
> Please met know, what was the intention behind mapping index -> id in the prompt, whether it was a typo (also considering that there is no `optionIndex` included in the prompt as far as I can see) or it was intentional, and whether including options ids in the prompt is okay. Perhaps the implementation should rely solely on option indexes and match by xpath could be removed?
> 
> ```
> "index": int, // the id corresponding to the optionIndex under the the select element.
> ```

nice catch. this is a legacy problem after we finished the HTML element tree.
can you also fix this in the `extract-action-claude3-sonnet.j2` file?

LawyZheng:
after you finish all suggestions, can you solve the conflict?
and we can start the testing.

ykeremy:
@LawyZheng we took over this PR, feel free to merge this on a new branch and then take over.

skyvern/forge/prompts/skyvern/extract-action.j2
```diff
@@ -13,14 +13,15 @@ Reply in JSON format with the following keys:
...
-        "id": int, // The id of the element to take action on. The id has to be one from the elements list
+        "id": str, // The id of the element to take action on. The id has to be one from the elements list
         "text": str, // Text for INPUT_TEXT action only
         "file_url": str, // The url of the file to upload if applicable. This field must be present for UPLOAD_FILE but can also be present for CLICK only if the click is to upload the file. It should be null otherwise.
         "download": bool, // Can only be true for CLICK actions. If true, the browser will trigger a download by clicking the element. If false, the browser will click the element without triggering a download.
         "option": {  // The option to select for SELECT_OPTION action only. null if not SELECT_OPTION action
             "label": str, // the label of the option if any. MAKE SURE YOU USE THIS LABEL TO SELECT THE OPTION. DO NOT PUT ANYTHING OTHER THAN A VALID OPTION LABEL HERE
-            "index": int, // the id corresponding to the optionIndex under the the select element.
+            "index": int, // the index corresponding to the index under the select element.
```
ykeremy, line 22:
```suggestion
            "index": int, // the index corresponding to the option index under the select element.
```

We add a specific `option index` field to the element tree. Let's keep this wording the same.

skyvern/forge/prompts/skyvern/extract-action.j2
ykeremy: Yes, we can remove this. I don't think we're even passing the option id to the LLM.

skyvern/webeye/actions/actions.py
```diff
@@ -93,9 +93,10 @@ class SelectOption(BaseModel):
     label: str | None
     value: str | None
     index: int | None
+    id: str | None
```
ykeremy, line 96: @LawyZheng let's remove this too

skyvern/webeye/actions/handler.py
ykeremy: Oh yeah this is a big bug 😮 We should stop using `id_to_xpath` dict and deprecate it IMO

PR #362, opened 2024-05-24
iFrames support

### LINKED ISSUE #8: Support interacting elements within iframes

alexzarbn:
## Problem
Crawling iframes is currently not supported, but to properly navigate some websites it is crucial to interact with iframes.

## Solution
- Crawl iframes by injecting `domUtils.js` into each iframe and building element tree by combining elements from the main page and iframes.
- When processing an action, resolve locator instances based on the frame the action has to take place (main page or a specific iframe).

_Note: the proposed implementation works with iframes one level deep, meaning iframes in iframes are not supported._

The PR is based on the #361 

Closes #8


skyvern/forge/prompts/skyvern/extract-action.j2
```diff
@@ -14,6 +14,7 @@ Reply in JSON format with the following keys:
...
         "id": str, // The id of the element to take action on. The id has to be one from the elements list
+        "frame": str, // It's either "main" or the id of the iframe element from the elements list
```
ykeremy, line 17: I'd like to keep the LLM unaware of frames. Can we create a mapping `unique_id -> frame` and decide which frame we're going to use based on the `unique_id` LLM returns?

Do you see any potential problems with the proposed approach ☝🏼 ?

skyvern/webeye/actions/actions.py
```diff
@@ -36,6 +36,7 @@ class Action(BaseModel):
 
 
 class WebAction(Action, abc.ABC):
+    frame: str | int
```
ykeremy, line 39: Same here, I think we only need to update the action handling logic. Specifically how playwright interacts with the page. 

We can build and store the `element_id_to_frame` mapping in the `ScrapedPage` object we build as a result of `scrape_web_unsafe`.

Then, we can pass that info to `resolve_locator` and trace back to the original frame there without changing the way LLM, Action objects work.

WDYT?

skyvern/webeye/scraper/domUtils.js
```diff
@@ -749,6 +750,11 @@ function buildTreeFromBody() {
         processElement(child, elementObj.id);
       });
       return elementObj;
+    } else if (element.tagName.toLowerCase() === 'iframe') {
+      let iframeElementObject = buildElementObject(element, true);
+
+      elements.push(iframeElementObject);
+      resultArray.push(iframeElementObject);
```
ykeremy, line 757: Should we skip the iframes here? 

```
function buildElementObject(element, interactable) {
```
Interactable here means that LLM can take an action such as CLICK/INPUT_TEXT/SELECT_OPTION etc on this specific element. Is that possible for this root iframe element here? Or are we making it possible via building an element tree for each of the iframes?

ykeremy (changes requested):
Thank you so much for looking into this! I was excited to even review it 🤩

I have some questions and concerns around some major points. Let me know if you want to hop on a call and discuss things. We can definitely resolve it async as well. Both works for me.

skyvern/webeye/actions/actions.py
alexzarbn: Sounds good! I have just pushed the update with id -> frame mapping. Please let me know, if everything is in order.

skyvern/forge/prompts/skyvern/extract-action.j2
alexzarbn: Not at all. It should also reduce any potential errors when LLM would not return the correct frame.

skyvern/webeye/scraper/domUtils.js
alexzarbn: The intention of including iframes in the elements list was to set unique id on the iframe for mapping later on and build element tree for each iframe. Not sure how to implement that without building the element object.

PR #370, opened 2024-05-27
fix null element execption

LawyZheng:


wintonzheng:
Why will the element be null? any example html?

LawyZheng:
> Why will the element be null? any example html?

I don't know. Only happen in the production. I can't reproduce it in my local. Just try to fix it according to the log.

wintonzheng:
@LawyZheng we need to figure out why this is happening. this change might hide the root cause

suchintan:
@wintonzheng it might be a good time to set up some sentry / browser logging so we can inspect console.logs

That being said, I THINK this is a pretty safe change?

skyvern/webeye/scraper/domUtils.js
```diff
@@ -714,6 +714,10 @@ function buildTreeFromBody() {
     }
   }
   function processElement(element, parentId) {
+    if (element === null) {
+      return;
```
wintonzheng, line 718: @LawyZheng 

let's console.log something here so that we can monitor issues like this from web console in the future

wintonzheng (approved):

skyvern/webeye/scraper/domUtils.js
LawyZheng: Sure, though there is no valuable info to log here, since the element is null and parentId might be null.

PR #380, opened 2024-05-28
choose the first valid email username

LawyZheng:


skyvern/forge/sdk/services/bitwarden.py
```diff
@@ -161,8 +167,18 @@ def get_secret_value_from_url(
...
+                return credentials[0]
+
+            # Choose multiple credentials according to the defined rule,
+            # if no cred matches the rule, return empty.
+            # TODO: For now hard code to choose the first valid email username
+            for cred in credentials:
+                if is_valid_email(cred.get(BitwardenConstants.USERNAME, "")):
+                    return cred
+
+            LOG.warning("No credential in Bitwarden matches the rule")
```
suchintan, line 180: In this case you should return the first match (not empty)

skyvern/forge/sdk/services/bitwarden.py
```diff
@@ -16,6 +17,11 @@
 LOG = structlog.get_logger()
 
 
+def is_valid_email(email: str) -> bool:
+    pattern = r"^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$"
```
wintonzheng, line 21: is this regex good enough? how did you come up with this one?

LawyZheng: It's generated by GPT 😆, and I just check the logic here. 
Regex like email is a common case that you can find a good answer through the Internet. So GPT should be good.

suchintan (approved):

PR #381, opened 2024-05-28
skip invalid actions during parsing

LawyZheng:


skyvern/webeye/actions/actions.py
```diff
@@ -133,108 +134,130 @@ class CompleteAction(DecisiveAction):
     data_extraction_goal: str | None = None
 
 
+def marshal_action(action: Dict[str, Any], data_extraction_goal: str | None = None) -> Action:
```
wintonzheng, line 137:
```suggestion
def parse_action(action: Dict[str, Any], data_extraction_goal: str | None = None) -> Action:
```
let's just use parse to keep the naming consistent to the existing `def parse_actions` function

wintonzheng (approved):

PR #417, opened 2024-06-05
Remove no executed result

LawyZheng:
| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit b70ed0ebcecb7385e6f961fe2fa84e710c9dbd85  | 
|--------|--------|

### Summary:
This PR updates the `agent_step` function to return cleaner action results by filtering out unsuccessful ones using `get_clean_detailed_output`.

**Key points**:
- Modify `agent_step` in `skyvern/forge/agent.py` to use `get_clean_detailed_output` from `DetailedAgentStepOutput`.
- Ensure only successful action results are returned.
- Applied to all return points in `agent_step` for failed and completed steps.

----

### DIFF
--- a/skyvern/forge/agent.py
+++ b/skyvern/forge/agent.py
@@ -18,7 +18,6 @@ from skyvern.exceptions import (
     InvalidWorkflowTaskURLState,
     MissingBrowserStatePage,
     StepTerminationError,
-    StepUnableToExecuteError,
     TaskNotFound,
 )
 from skyvern.forge import app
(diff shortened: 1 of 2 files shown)


LawyZheng:
> The PR looks good, my question to you is about the approach:
> 
> Do we even need to store the actions that we didn't even try executing in the `actions_and_results`? We already store all actions in `DetailedAgentStepOutput.actions`, so there'd be no information loss if we just didn't have those actions in `DetailedAgentStepOutput.actions_and_results` in the first place. WDYT?

Do you mean we don't need the `DetailedAgentStepOutput.actions_and_results`? To be honest, I don't know the answer. We got `DetailedAgentStepOutput.actions_and_results`, `DetailedAgentStepOutput.actions`, `DetailedAgentStepOutput.action_results` and we also got `AgentStepOutput.action_results`, `AgentStepOutput.actions_and_results`.
Quite parts of them seem to be duplicated. I don't know why we created these in the first place. But I don't think we should remove these right now. Maybe we could do it when we refactor the `StepOutput` part. I don't want to create any unexpected bugs in this PR.

PR #557, opened 2024-07-05
exclude aria-hidden elements

LawyZheng:
| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit 621daf3132f43c516c9e66762067cc2b6970fe47  | 
|--------|--------|

### Summary:
Updated `isElementVisible` in `skyvern/webeye/scraper/domUtils.js` to exclude elements with `aria-hidden='true'`.

**Key points**:
- Updated `skyvern/webeye/scraper/domUtils.js`.
- Modified `isElementVisible` function to check for `aria-hidden` attribute.
- Elements with `aria-hidden='true'` are now considered not visible.

----

### DIFF
--- a/skyvern/webeye/scraper/domUtils.js
+++ b/skyvern/webeye/scraper/domUtils.js
@@ -223,6 +223,11 @@ function isElementVisible(element) {
     return false;
   }
 
+  if (element.hasAttribute("aria-hidden")) {
+    const ariaHidden = element.getAttribute("aria-hidden").toLowerCase();
+    return ariaHidden === "false";
+  }
+
   const style = getElementComputedStyle(element);
   if (!style) return true;
   if (style.display === "contents") {


skyvern/webeye/scraper/domUtils.js
```diff
@@ -223,6 +223,11 @@ function isElementVisible(element) {
     return false;
   }
 
+  if (element.hasAttribute("aria-hidden")) {
+    const ariaHidden = element.getAttribute("aria-hidden").toLowerCase();
+    return ariaHidden === "false";
```
LawyZheng, line 228: except aria-hiden="False" or aria-hiden="false", all aria-hidden values should be considered as hidden, like `aria-hidden`, `aria-hidden=""`, `aria-hidden="true"`, `aria-hidden="True"`

ykeremy (changes requested):
there is already a isHidden function in this file, we should update it instead.

LawyZheng:
> there is already a isHidden function in this file, we should update it instead.

we only use the isHidden function for option element, should I enable it to use for all elements ?

LawyZheng:
Close this PR since it's still controversial to make this change.

PR #604, opened 2024-07-15
Skip if selected

LawyZheng:
| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit ecd7cef40b36ec8c3bb53cf5b94c793a7fca0e1d  | 
|--------|--------|

### Summary:
Refactored `handle_select_option_action` to check for already selected options in `normal_select` and updated `getSelectOptions` to return the selected option's text.

**Key points**:
- Updated `handle_select_option_action` in `skyvern/webeye/actions/handler.py` to remove the check for already selected options.
- Moved the check for already selected options to `normal_select` in `skyvern/webeye/actions/handler.py`.
- Modified `getSelectOptions` in `skyvern/webeye/scraper/domUtils.js` to return the selected option's text along with the options list.
- Added handling for `selected` attribute in `buildTreeFromBody` in `skyvern/webeye/scraper/domUtils.js`.

----

### DIFF
--- a/skyvern/webeye/actions/handler.py
+++ b/skyvern/webeye/actions/handler.py
@@ -424,7 +424,6 @@ async def handle_select_option_action(
 ) -> list[ActionResult]:
     dom = DomUtil(scraped_page, page)
     skyvern_element = await dom.get_skyvern_element_by_id(action.element_id)
-    locator = skyvern_element.locator
 
     tag_name = skyvern_element.get_tag_name()
     element_dict = scraped_page.id_to_element_dict[action.element_id]
@@ -830,7 +829,7 @@ async def normal_select(
 
     action_result: List[ActionResult] = []
     is_success = False
-    locator = skyvern_element.locator
+    locator = skyvern_element.get_locator()
 
     try:
         await locator.click(
--- a/skyvern/webeye/actions/handler.py
+++ b/skyvern/webeye/actions/handler.py
@@ -583,13 +583,6 @@ async def handle_select_option_action(
         click_action = ClickAction(element_id=action.element_id)
         return await chain_click(task, scraped_page, page, click_action, skyvern_element)
 
-    try:
-        current_text = await locator.input_value()
-        if current_text == action.option.label or current_text == action.option.value:
-            return [ActionSuccess()]
-    except Exception:
-        LOG.info("failed to confirm if the select option has been done, force to take the action again.")
-
     return await normal_select(action=action, skyvern_element=skyvern_element)
 
 
@@ -828,6 +821,13 @@ async def normal_select(
     action: actions.SelectOptionAction,
     skyvern_element: SkyvernElement,
 ) -> List[ActionResult]:
+    try:
+        current_text = await skyvern_element.get_attr("selected")
+        if current_text == action.option.label or current_text == action.option.value:
+            return [ActionSuccess()]
+    except Exception:
+        LOG.info("failed to confirm if the select option has been done, force to take the action again.")
+
     action_result: List[ActionResult] = []
     is_success = False
     locator = skyvern_element.locator
--- a/skyvern/webeye/scraper/domUtils.js
+++ b/skyvern/webeye/scraper/domUtils.js
@@ -606,8 +606,8 @@ function getSelectOptions(element) {
     });
   }
 
-  const selectedOption = element.querySelector("option:checked")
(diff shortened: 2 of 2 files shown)


skyvern/webeye/actions/handler.py
```diff
@@ -828,9 +820,16 @@ async def normal_select(
     action: actions.SelectOptionAction,
     skyvern_element: SkyvernElement,
 ) -> List[ActionResult]:
+    try:
+        current_text = await skyvern_element.get_attr("selected")
```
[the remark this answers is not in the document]
LawyZheng: A good point here. Such as if the previous action caused the select value reset, the logic might be wrong. But maybe we can re-do the select in the next step. So it's not a blocker here.

skyvern/webeye/actions/handler.py
```diff
@@ -828,9 +820,16 @@ async def normal_select(
...
+        if current_text == action.option.label or current_text == action.option.value:
+            return [ActionSuccess()]
+    except Exception:
+        LOG.info("failed to confirm if the select option has been done, force to take the action again.")
+
     action_result: List[ActionResult] = []
     is_success = False
-    locator = skyvern_element.locator
+    locator = skyvern_element.get_locator()
```
suchintan, line 832: whoa seems like a big bug

skyvern/webeye/actions/handler.py
```diff
@@ -828,9 +820,16 @@ async def normal_select(
     action: actions.SelectOptionAction,
     skyvern_element: SkyvernElement,
 ) -> List[ActionResult]:
+    try:
```
suchintan, line 823: I don't think this is a good change. How can the LLM "unselect" an option? Here doing select option on an already selected option would unselect it before, but now that can never be the case right?

skyvern/webeye/actions/handler.py
LawyZheng: ? What do you mean? I don't follow

suchintan: Sorry for the confusion! Did the old.locator implementation work? If not that's a huge bug! Nice find.

skyvern/webeye/actions/handler.py
LawyZheng: What do you mean unselect before select? I don't understand.

skyvern/webeye/actions/handler.py
LawyZheng: Oh both of them could work. I just refactor it into the same format.

skyvern/webeye/actions/handler.py
LawyZheng: If llm want to unselect sth. He can return the value represented the "unselect" right? What the problem here?

skyvern/webeye/scraper/domUtils.js
```diff
@@ -605,7 +605,13 @@ function getSelectOptions(element) {
...
     });
   }
-  return selectOptions;
+
+  const selectedOption = element.querySelector("option:checked");
+  if (!selectedOption) {
+    return [selectOptions, ""];
+  }
+
+  return [selectOptions, removeMultipleSpaces(selectedOption.textContent)];
```
wintonzheng, line 614: is the selected value always textContent? could it be the "value" attribute of the element?

skyvern/webeye/actions/handler.py
```diff
@@ -828,9 +820,16 @@ async def normal_select(
...
+        current_text = await skyvern_element.get_attr("selected")
+        if current_text == action.option.label or current_text == action.option.value:
+            return [ActionSuccess()]
+    except Exception:
+        LOG.info("failed to confirm if the select option has been done, force to take the action again.")
```
wintonzheng, line 828: why do we say "force to take the action again"? we're not retrying here, are we?

wintonzheng (approved):

skyvern/webeye/scraper/domUtils.js
LawyZheng: because when we parse the option. We only use the textContext to represent the value. I think we should be consistent with the logic we do in the option parsing.

skyvern/webeye/actions/handler.py
LawyZheng: em. if the skip logic failed. we try to select again!

PR #1079, opened 2024-10-29
Usage of Ollama and local models

loktar00:
This is terrible code and I know it, I don't expect this to be merged but this is here in the hopes someone else will be able to get it working a bit better. 

Full caveat, this "works" but every task has actually failed. For the local models being used the prompts would definitely need to be adjusted, however I did get the correct answer on 2 occasions in the logs.

### Summary:

* Split vision and language models, this may not have been 100% necessary but it accounts for ollama not allowing parallel requests. In addition the vision models in Ollama aren't as good as language.
* Adds some more env variables all related to Ollama
* Disables cost calculations when using ollama since we don't care or need that for local
* Added options for resolution in the configs, just makes the images a bit lighter for local vision models.
* Modified parsing for local models to target ` ```json` 

### Things that exclude it from being merged and would love help with
* Added files directly to the volumes in the `docker-compose`
* Added a bunch of logging for debugging would need to be removed
* Tis messy, I wont lie i used cursor a bit, I'm not an experienced Python Dev, I'm a JS guy ;)
* I DID NOT TEST EXISTING FUNCTIONALITY / MODELS so there could be breaking changes.

----

> [!IMPORTANT]
> Adds support for Ollama models, modifies configurations, and includes debugging logs and Docker setup changes.
> 
>   - **Behavior**:
>     - Split vision and language models to handle Ollama's non-parallel request limitation in `api_handler_factory.py`.
>     - Modified parsing for local models to target ` ```json` in `api_handler_factory.py`.
>     - Disables cost calculations for Ollama models in `models.py`.
>   - **Configuration**:
>     - Added Ollama-specific settings in `config.py` and `config_registry.py`.
>     - Updated `docker-compose.yml` to include Ollama environment variables and volume mounts.
>   - **Misc**:
>     - Added extensive logging for debugging in `api_handler_factory.py`.
>     - Directly added files to Docker volumes in `docker-compose.yml`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=Skyvern-AI%2Fskyvern&utm_source=github&utm_medium=referral)<sup> for 3e3f4de4eb8558bc7dbf6d1cb4bd84f85d3b0290. It will automatically update as commits are pushed.</sup>


suchintan:
haha I love to see terrible code (as a v1!!)

docker-compose.yml
```diff
@@ -23,29 +21,35 @@ services:
...
-      # - ENABLE_AZURE=true
-      # - LLM_KEY=AZURE_OPENAI
-      # - AZURE_DEPLOYMENT=<your_azure_deployment>
-      # - AZURE_API_KEY=<your_azure_api_key>
-      # - AZURE_API_BASE=<your_azure_api_base>
-      # - AZURE_API_VERSION=<your_azure_api_version>
+      # Enable Ollama
+      - ENABLE_OLLAMA=true
+      - LLM_KEY=OLLAMA_TEXT
+      - OLLAMA_API_BASE=http://192.168.1.3:11434
```
DominicTWHV, line 41: Add option to persist model in memory

curl http://localhost:11434/api/generate -d '{
  "model": "<model>",
  "keep_alive": <keepalive>
}'

art3miz18:
@loktar00 hey, is this compactible with llama vision models ?

loktar00:
> @loktar00 hey, is this compactible with llama vision models ?

Should be fine with it for the first step, it switches between models, vision then it goes to a language model. Only because Ollama doesn't support both at once, unless the Vision Support for Llama works now.

DominicTWHV:
Does skyvern not use opencv? Their public version with the box drawing looks like what opencv does. After that, it passes the images to a vision model I think for analysis.

DominicTWHV:
You can rapidly load and unload ollama models via the api endpoint as suggested above, use keepalive 0 to immediately unload said model.

DominicTWHV:
Btw, check out this pull:
https://github.com/Skyvern-AI/skyvern/pull/1160

Seems like the OP has issues with ollama based vision models for some reason

Prakashmaheshwaran:
ollama support added #2283 ✨✌️

DominicTWHV:
That is sick

PR #1287, opened 2024-11-28
Remove version from docker compose file as it's deprecated

Kumala3:


wintonzheng:
Thank you. i'll merge it

Issue #1346, opened 2024-12-07
Downloading PDF files freeze the agent


n-sviridenko:
When the agent clicks a PDF file link to download it, it opens PDF in a new tab (due to Chromium's default PDF viewer capability), this makes the agent losing the initial page (where there are other links to click) and not coming back. And then it just hangs on that PDF page forever.

n-sviridenko:
In order to reproduce it, create a page with lots of links to PDF files and tasks Skyvern to click all of them.
It'll only visit the first PDF.

n-sviridenko:
Btw. we also need to download PDFs instead of just viewing them.

n-sviridenko:
Was also trying to see how downloading invoices is currently implemented https://docs.skyvern.com/getting-started/skyvern-in-action#log-into-a-portal-and-download-invoices

n-sviridenko:
I assume implementation is:
1. subscribing to opened pages (before `await locator.click`)
2. if a new one just got opened
3. try fetching it (HEAD request, but w/ cookies of the current session, e.g. via calling `fetch`)
4. if content type from headers is PDF, then come back to prev page and download programmatically using the same URL and cookies from current session

Place to change:
https://github.com/Skyvern-AI/skyvern/blob/92d94954c4dab976ff796caa4baf65d53990ea74/skyvern/webeye/actions/handler.py#L390

suchintan:
We actually have this implemented in Skyvern cloud similar to what you did here: https://github.com/Skyvern-AI/skyvern/pull/1349/files but I think we forgot to bring it to the open source repo

@LawyZheng can you please review https://github.com/Skyvern-AI/skyvern/pull/1349/files and make any augmentations? We should open source our PDF handling logic

LawyZheng:
i just left some comments a few minutes ago.

suchintan:
@LawyZheng let's get our closed source browser stuff ported over to the open source version

LawyZheng:
Fixed in https://github.com/Skyvern-AI/skyvern/pull/1363

PR #1349, opened 2024-12-07
Fix PDF downloads (option 2)

n-sviridenko:
Fixing this issue: https://github.com/Skyvern-AI/skyvern/issues/1346

### DIFF
--- a/skyvern/webeye/browser_factory.py
+++ b/skyvern/webeye/browser_factory.py
@@ -28,6 +28,8 @@ from skyvern.forge.sdk.core.skyvern_context import current
 from skyvern.forge.sdk.schemas.tasks import ProxyLocation
 from skyvern.webeye.utils.page import SkyvernFrame
 
+import json
+
 LOG = structlog.get_logger()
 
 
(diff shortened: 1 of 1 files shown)


skyvern/webeye/browser_factory.py
```diff
@@ -66,40 +67,37 @@ async def listen_to_download(download: Download) -> None:
...
-                    file_path.rename(str(file_path) + suffix)
-                    return
-                # TODO: maybe should try to parse it from URL response
+                    suffix = download_url_suffix
+
+                name = Path(file_path).name
+                if suffix:
+                    name += suffix
+                new_file_path = download_dir / name
+                await download.save_as(new_file_path)
```
LawyZheng, line 100: what's the specific bug in the original logic?

skyvern/webeye/browser_factory.py
```diff
@@ -66,40 +67,37 @@ async def listen_to_download(download: Download) -> None:
...
                     )
-                    file_path.rename(str(file_path) + suffix)
-                    return
-                # TODO: maybe should try to parse it from URL response
+                    suffix = download_url_suffix
+
+                name = Path(file_path).name
+                if suffix:
+                    name += suffix
+                new_file_path = download_dir / name
```
LawyZheng, line 99: what it means `download / name`? they're both str, not int or float.

skyvern/webeye/browser_factory.py
```diff
@@ -146,8 +144,21 @@ def build_browser_args() -> dict[str, Any]:
...
+            },
+        }
+        user_data_dir = make_temp_directory(prefix="skyvern_browser_")
+        preference_dir = os.path.join(user_data_dir, "Default")
+        os.makedirs(preference_dir, mode=0o777, exist_ok=True)
+        with open(preference_dir + "/Preferences", "w") as f:
+            json.dump(preference, f)
+
         return {
-            "user_data_dir": make_temp_directory(prefix="skyvern_browser_"),
```
LawyZheng, line 150: adding a new key like `downloads_path` here is able to change the default download directory. no need to change the logic in the `set_download_file_listener` function

skyvern/webeye/browser_factory.py
n-sviridenko: because of this error downloads were not saved @LawyZheng , as `suffix` was `None` @LawyZheng

LawyZheng: as long as we set the download_paths in the browser args. i don't think we need this change. 
https://github.com/Skyvern-AI/skyvern/blob/583075899fc9938d63ca37f6c1005e3b7306de4a/skyvern/webeye/browser_factory.py#L294

https://github.com/Skyvern-AI/skyvern/blob/583075899fc9938d63ca37f6c1005e3b7306de4a/skyvern/webeye/browser_factory.py#L316

n-sviridenko:
Then the agent that sets download_triggered needs to check them as well. As
when I tested it, it failed

PR #1699, opened 2025-02-01
Update Dockercompose with more info

suchintan:
> [!IMPORTANT]
> Enhance `docker-compose.yml` with new `LLM_KEY` and detailed DNS routing instructions for self-hosting.
> 
>   - **Environment Variables**:
>     - Added `LLM_KEY=OPENAI_GPT4O` to `skyvern` service.
>     - Updated comment for `VITE_SKYVERN_API_KEY` in `skyvern-ui` service to guide users on obtaining the key.
>   - **DNS Routing Instructions**:
>     - Added detailed comments for DNS routing when self-hosting, including routes for API, UI, and artifact API in `skyvern-ui` service.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=Skyvern-AI%2Fskyvern&utm_source=github&utm_medium=referral)<sup> for 6a4e9f267ac5e58f106b92c192814bd88eec1eb0. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -97,7 +97,7 @@ services:
     # If you're self-hosting this behind a dns, you'll want to set:
     #   A route for the API: api.yourdomain.com -> localhost:8000
     #   A route for the UI: yourdomain.com -> localhost:8080
-    #   A route for the artifact API: artifact.yourdomain.com -> localhost:9090
+    #   A route for the artifact API: artifact.yourdomain.com -> localhost:9090 (maybe not needed)
       - VITE_WSS_BASE_URL=ws://localhost:8000/api/v1
     #   - VITE_ARTIFACT_API_BASE_URL=http://localhost:9090
     #   - VITE_API_BASE_URL=http://localhost:8000/api/v1
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -36,6 +36,7 @@ services:
       - DATABASE_STRING=postgresql+psycopg://skyvern:skyvern@postgres:5432/skyvern
       - BROWSER_TYPE=chromium-headful
       - ENABLE_OPENAI=true
+      - LLM_KEY=OPENAI_GPT4O
       - OPENAI_API_KEY=<your_openai_key>
       # If you want to use other LLM provider, like azure and anthropic:
       # - ENABLE_ANTHROPIC=true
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -93,9 +94,14 @@ services:
...
     # you need to change the host in VITE_WSS_BASE_URL and VITE_API_BASE_URL to match your server ip
+    # If you're self-hosting this behind a dns, you'll want to set:
+    #   A route for the API: api.yourdomain.com -> localhost:8000
+    #   A route for the UI: yourdomain.com -> localhost:8080
+    #   A route for the artifact API: artifact.yourdomain.com -> localhost:9090 (maybe not needed)
       - VITE_WSS_BASE_URL=ws://localhost:8000/api/v1
+    #   - VITE_ARTIFACT_API_BASE_URL=http://localhost:9090
     #   - VITE_API_BASE_URL=http://localhost:8000/api/v1
-    #   - VITE_SKYVERN_API_KEY=
+    #   - VITE_SKYVERN_API_KEY=<get this from "settings" in the Skyvern UI>
```
wintonzheng, line 104: https://github.com/Skyvern-AI/skyvern/blob/204972e2253d31eddb71ed26a22903cd8258598a/entrypoint-skyvernui.sh#L7

currently we use the api key in `.streamlit/secrets.toml` so this is not a required step here.

we should migrate to a better solution - use an .env file

docker-compose.yml
```diff
@@ -93,9 +94,14 @@ services:
     environment:
     # if you want to run skyvern on a remote server,
     # you need to change the host in VITE_WSS_BASE_URL and VITE_API_BASE_URL to match your server ip
+    # If you're self-hosting this behind a dns, you'll want to set:
+    #   A route for the API: api.yourdomain.com -> localhost:8000
+    #   A route for the UI: yourdomain.com -> localhost:8080
+    #   A route for the artifact API: artifact.yourdomain.com -> localhost:9090 (maybe not needed)
```
wintonzheng, line 100: @msalihaltun is the 9090 port still used? if not we should remove it.

docker-compose.yml
```diff
@@ -93,9 +94,14 @@ services:
     environment:
     # if you want to run skyvern on a remote server,
     # you need to change the host in VITE_WSS_BASE_URL and VITE_API_BASE_URL to match your server ip
+    # If you're self-hosting this behind a dns, you'll want to set:
+    #   A route for the API: api.yourdomain.com -> localhost:8000
```
wintonzheng, line 98: this is a bit confusing to me. when I set up dns routing, I won't use localhost. what we want say here is to point their api.yourdomain.com to the hosting service's dns or ip, port 8000

suchintan: that's correct, and I'm not sure if this commetn is that confusing. I'm OK to let our users say that because only our very very technical users are going to be reading this comment and leveraging it.

docker-compose.yml
suchintan: it's used for the artifacts endpoint

docker-compose.yml
suchintan: not high priority imo!

wintonzheng (approved):

PR #1784, opened 2025-02-18
update element hashing logic - subscribe only. otherwise, any element update would break element hashing

wintonzheng:


skyvern/webeye/scraper/scraper.py
```diff
@@ -156,9 +156,12 @@ def json_to_html(element: dict, need_skyvern_attrs: bool = True) -> str:
 
 
 def clean_element_before_hashing(element: dict) -> dict:
-    element_copy = copy.deepcopy(element)
-    element_copy.pop("id", None)
-    element_copy.pop("rect", None)
+    element_copy = {}
+    # pick up these element keys: frame, tagName, attributes, beforePseudoText, text, afterPseudoText, children
+    for key in ["frame", "tagName", "attributes", "beforePseudoText", "text", "afterPseudoText", "children"]:
```
LawyZheng, line 161: why should we keep `frame` attribute?

wintonzheng: i'll remove.

suchintan (approved):

deveshruttala-edu:

We’d love your feedback!

PR #2221, opened 2025-04-22
Poc/forward user clicks to running browser

jomido:
Hello, Skyvern!

## What 

POC for sending user click events to the running (headful/cdp) browser.

## Why 

Because it's neat/useful.

### DX

I could not get a development flow that had both of these reqs:
- showed the live stream of the browser, in-app
- fast iteration for code updates 

The local development, per the instructions in the `README.md`, created a setup where the stream of the browser was missing. This is kind of important for iteration speed. Having it not be there is a non-starter.

The docker compose scenario, as outlined in the `README.md`, created a setup where local code changes (to the frontend) were not honored. Nor could I run `vite` in dev mode (`npm run dev`), as `npm run start` was baked into the cached image.

So here is what I used:

```yaml
services:
  postgres:
    image: postgres:14-alpine
    restart: always
    volumes:
      - ./postgres-data:/var/lib/postgresql/data
    environment:
      - PGDATA=/var/lib/postgresql/data/pgdata
      - POSTGRES_USER=skyvern
      - POSTGRES_PASSWORD=skyvern
      - POSTGRES_DB=skyvern
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -U skyvern"]
      interval: 5s
      timeout: 5s
      retries: 5

  skyvern:
    image: public.ecr.aws/skyvern/skyvern:latest
    restart: on-failure
    ports:
      - 8000:8000
    volumes:
      - ./artifacts:/data/artifacts
      - ./videos:/data/videos
      - ./har:/data/har
      - ./log:/data/log
      - ./.streamlit:/app/.streamlit
      - ./skyvern:/app/skyvern
      - ./alembic:/app/alembic
    environment:
      - DATABASE_STRING=postgresql+psycopg://skyvern:skyvern@postgres:5432/skyvern
      - BROWSER_TYPE=chromium-headful
      - ENABLE_OPENAI=true
      - LLM_KEY=OPENAI_GPT4O
      - OPENAI_API_KEY=<your-llm-api-key>
    depends_on:
      postgres:
        condition: service_healthy
    healthcheck:
      test: ["CMD", "test", "-f", "/app/.streamlit/secrets.toml"]
      interval: 5s
      timeout: 5s
      retries: 5

  skyvern-ui:
    image: public.ecr.aws/skyvern/skyvern-ui:latest
    restart: on-failure
    ports:
      - 8080:8080
      - 9090:9090
    volumes:
      - ./artifacts:/data/artifacts
      - ./videos:/data/videos
      - ./har:/data/har
      - ./.streamlit:/app/.streamlit
      - ./skyvern-frontend/src:/app/src
    environment:
      - VITE_WSS_BASE_URL=ws://localhost:8000/api/v1
      - VITE_SKYVERN_API_KEY=<copy-from-streamlit/secrets.toml>
    depends_on:
      skyvern:
        condition: service_healthy
    command: ./node_modules/vite/bin/vite.js --host=0.0.0.0 & npm run run-artifact-server
```

This took me a while to figure out (multiple sit-downs). There were other blockers, but I've banished them from my working memory. On to the juice!

### In Short

- capture 0-1 coords over streamed browser image; save 'em
  - account for CSS transforms (`object-fit`)
  - account for browser chrome (address bar, buttons, etc.) 
- when image is clicked, send a WebSocket message to the backend
  - allow existing backend WS to receive messages (was only sending them)
- use existing `BrowserManager` to send Playwright message

### Again, with Images

Run a workflow:

 

Click on the browser to zoom in:

Click around the browser. Circular debug divs are added. They'll disappear 10 seconds after insertion:

## Notes 

- not intended to merge; just a POC
- keyboard events look trivial to add
- looks slow/laggy; but stream is slow; pretty sure events themselves are "fast enough"
- current WS loop is non-optimal
- I hacked the `BrowserManager` API - maybe there's some preferred way to use it - not sure

### DIFF
--- a/skyvern-frontend/src/components/ZoomableImage.tsx
+++ b/skyvern-frontend/src/components/ZoomableImage.tsx
@@ -1,11 +1,39 @@
-import { useEffect, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import clsx from "clsx";
+import * as apiTypes from "@/api/types";
 
-type HTMLImageElementProps = React.ComponentProps<"img">;
+interface ZoomableImageProps extends React.ComponentProps<"img"> {
+  captureClick?: (pos: apiTypes.Position) => void;
+}
 
-function ZoomableImage(props: HTMLImageElementProps) {
+function ZoomableImage({ captureClick, ...props }: ZoomableImageProps) {
   const [modalOpen, setModalOpen] = useState(false);
   const [zoom, setZoom] = useState(false);
+  const imgRef = useRef<HTMLImageElement>(null);
+  const [position, setPosition] = useState({ x: 0, y: 0 });
+
+  const classes = {
+    img: clsx("cursor-zoom-in object-contain", props.className),
+    modalCloseButton:
+      "absolute right-4 top-4 cursor-pointer text-4xl text-white",
+    modalContainer: clsx(
+      "fixed inset-0 z-50 flex justify-center overflow-auto bg-black bg-opacity-75 p-16",
+      {
+        "items-center": !zoom,
+        "items-baseline": zoom,
+      },
+    ),
+    modalImg: captureClick
+      ? clsx(
+          "m-0 h-full max-h-full min-h-full w-full max-w-full object-contain",
+        )
+      : clsx({
+          "m-0 h-full max-h-full min-h-full w-full max-w-full cursor-zoom-in object-contain":
+            !zoom,
+          "m-0 ml-auto mr-auto max-h-none min-h-full max-w-none cursor-zoom-out object-contain":
+            zoom,
+        }),
+  };
 
   const openModal = () => {
     setZoom(false);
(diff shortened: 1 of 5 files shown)


skyvern-frontend/src/components/ZoomableImage.tsx
```diff
@@ -16,6 +44,75 @@ function ZoomableImage(props: HTMLImageElementProps) {
...
+      bottom: top + renderedHeight,
+    };
+  };
+
+  /**
+   * NOTE(jdo): These are constant offsets to make the click events happen at
+   * the correct location in the running browser. I believe the `y` component
+   * is the height of the browser chrome (address bar, buttons, etc.).
+   */
+  const magicConstants = {
```
suchintan, line 95: these should be in the backend, not front-end

skyvern/forge/sdk/routes/streaming.py
```diff
@@ -220,6 +222,61 @@ async def workflow_run_streaming(
...
+
+            """
+            TODO(jdo): can use an automated checker against the message type,
+            smth like typeguard; manual parsing is tedious/error prone; this
+            will require specifying the types on the backend, and keeping them
+            in sync with the TypeScript types on the frontend (there may be some
+            ways to automate that bit, too)
+            """
+            try:
+                while True:
```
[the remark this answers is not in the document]
suchintan: @jomido what do you think of this?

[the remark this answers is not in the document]
wintonzheng: is this while loop going to block all the logic above that streams the workflow run status and the encoded screenshot?

skyvern/forge/sdk/routes/streaming.py
```diff
@@ -220,6 +222,61 @@ async def workflow_run_streaming(
...
+                    async with asyncio.timeout(0.1):
```
wintonzheng, line 235: why are we timing out after 0.1 seconds?

if client_message is received but it takes longer than 0.1 seconds to do `await user_protocol.click(workflow_run_id, x, y)`, what will happen?

wintonzheng: i think what we want is probably handling the "click event" completely asynchronously by using [asyncio.create_task](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task)

wintonzheng: I agree message receiving should take less than 1 s to do, thus `0.1` sort of makes sense to me now but all the event handling should probably be handled in a separate event loop

wintonzheng: i imagine there will be 2 event loops, the main one (this while loop) collects the messages and put all the messages in a queue, and another separate event loop that processes messages in the queue.

suchintan:
Interesting. I just tested it and I'm not sure if it's working as intended

https://www.loom.com/share/2cde50774212441ea3d3f2f55469c810

1. The click -> interactoin feedback loop is a little longer than expected
2. The image refreshing creates a weird experience haha
3. I love the circles you make for interactions
4. can't refresh the page

skyvern/forge/sdk/routes/streaming.py
wintonzheng: another thought: this bi direction communication becomes complex.

shall we even just create a new streaming endpoint for event handling? this part of logic is unrelated to sending the screenshot to the client at all.

skyvern/forge/sdk/routes/streaming.py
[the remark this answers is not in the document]
jomido: > @jomido what do you think of this?

I mentioned that in a comment. `ellipses-dev` probably lifted from that.

> is this while loop going to block all the logic above that streams the workflow run status and the encoded screenshot?

No. There's a timeout of 0.1. Which can be lower. Having this embedded here is not ideal. I mentioned that the WS loop is not ideal in the PR description.

skyvern-frontend/src/components/ZoomableImage.tsx
jomido: Agreed.

skyvern/forge/sdk/routes/streaming.py
jomido: Yes this is complex. It is not ideal. I was under the impression that this effort was to be quick and dirty.

For starters, I'd investigate putting the stream into WebRTC.

For WebSockets, I'd reify a `Client` for each connection and pump messages bi-directionally via an `asyncio.Queue`.

jomido:
> The click -> interactoin feedback loop is a little longer than expected

It's because the current streaming mechanism is slow. I didn't mess with that.

> The image refreshing creates a weird experience haha

You should see it on my slow connection. :)

> I love the circles you make for interactions

`ellipses-dev` suggested I put this behind a debug flag. I agree, but this is quick and dirty, as requested.

> can't refresh the page

No. You shouldn't stream the whole browser, including it's address bar and UI buttons. Just the page contents. Then build your own address bar and other UI to mimic a browser, locally, in the web app.

skyvern-frontend/src/components/ZoomableImage.tsx
jomido: In addition, there's a good chance these constants won't be in play if the stream includes only page contents (not browser address bar, UI, etc.)

Prakashmaheshwaran:
One of most important feature i was tunneling a local debugger browser to my hosted skyvern, just bcaz i could click and enter few things manually or captcha, would be great if i could enter values too.

PR #2380, opened 2025-05-19
Added support for python 3.12 & 3.13

Prakashmaheshwaran:


skyvern/cli/commands.py
```diff
@@ -3,69 +3,910 @@
...
+    except Exception:
+        db_status = "❌ Not connected"
+    
+    # Check if server is running (port 8000)
+    server_status = "⏳ Checking..."
+    try:
+        import socket
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        s.settimeout(0.5)
+        s.connect(("localhost", 8000))
```
LawyZheng, line 332: can we also check the status from skyvern cloud?

skyvern/cli/commands.py
```diff
@@ -3,69 +3,910 @@
...
+    except:
+        server_status = "❌ Not running"
+    
+    # Check if UI is running (port 8080)
+    ui_status = "⏳ Checking..."
+    try:
+        import socket
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        s.settimeout(0.5)
+        s.connect(("localhost", 8080))
```
LawyZheng, line 344: same here

skyvern/cli/commands.py
Prakashmaheshwaran: No this particularly only works for localhost, but can add cloud version also if there is a API endpoint for it

wintonzheng (approved):

PR #2515, opened 2025-05-29
Fix broken pre-commit checks (isort/ruff formatting)

asherf:
https://github.com/Skyvern-AI/skyvern/actions/runs/15327550672/job/43125835803#step:10:217


asherf:
@wintonzheng @suchintan I suggest enabling https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/managing-a-branch-protection-rule to prevent CI from breaking on master.

suchintan (approved):

asherf:
> hmm we have the protection rule. not sure why it didn't block https://github.com/Skyvern-AI/skyvern/pull/2510

I think you have a branch protection rule, but you need to enable specific CI checks that are required to pass in order for it to block if there are CI failures

PR #2519, opened 2025-05-29
chore: remove setup.sh script as CLI skyvern quickstart does the same…

Prakashmaheshwaran:
chore: remove setup.sh script as CLI skyvern quickstart does the same and .sh is not maintained too


asherf:
FWIW - I think this is a good idea, however, to use the skyvern CLI, one must run `pip install skyvern` which install the latest version from pypi and not the local version checked out (assuming someone has the repo locally)
the setup.sh script is still needed for this scenario I think...
so I think some documentation updates are needed in order to provide guidance on how to setup local dev environment in the absence of if `setup.sh`.
I am sure there is a way to executer the "local" version of the skyvern CLI in that case, but I can't seem to find any guidance about it...
maybe I am missing something.

Prakashmaheshwaran:
added  #2523

wintonzheng (approved):

PR #2531, opened 2025-05-29
Fix Claude 4 model names to use specific versions instead of -latest suffix

tigershen23:
Updates Claude 4 model configurations to use specific version identifiers instead of unsupported -latest suffix. According to [Anthropic's model documentation](https://docs.anthropic.com/en/docs/about-claude/models/overview#model-names), the `-latest` suffix is not supported by the Claude 4 family models. Using the unsupported suffix leads to 404 API errors


wintonzheng (approved):

wintonzheng:
Thanks for the fix

PR #2601, opened 2025-06-05
Add run lifecycle timestamps - improvments

Prakashmaheshwaran:


alembic/versions/2025_06_02_0050-add_run_timestamps.py
```diff
@@ -0,0 +1,44 @@
...
+from alembic import op
+import sqlalchemy as sa
+
+revision: str = "add_run_timestamps"
+down_revision: Union[str, None] = "babaa7307e8a"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
```
suchintan, line 20: You're gonna have to regenerate this when you go to cloud

scripts/test_persistent_browsers.py
```diff
@@ -2,6 +2,10 @@
 import os
 from typing import Any, Optional, cast
 
+import pytest
```
suchintan, line 5: Why do you need this?

suchintan (approved):
Generally looks ok. Please test all 3 run modes

alembic/versions/2025_06_02_0050-add_run_timestamps.py
Prakashmaheshwaran: yep

deveshruttala-edu:

We’d love your feedback!

PR #2613, opened 2025-06-05
Don't allow None for org id (partial mypy only change) more to come

asherf:
it seems that we should allow this.
partial change since in a lot of places org id is kwarg with a default value.
will make more changes down the line.

this is a mypy only change so should be very safe.

----

> [!IMPORTANT]
> Enforce non-null `organization_id` in function signatures and class attributes across multiple files for mypy type checking.
> 
>   - **Function Signatures**:
>     - Change `organization_id` from `str | None` to `str` in `create_task()` in `agent.py`.
>     - Change `organization_id` from `str | None` to `str` in `validate_task_execution()` in `agent_functions.py`.
>     - Change `organization_id` from `str | None` to `str` in `get_task_steps()` in `client.py`.
>   - **Class Attributes**:
>     - Change `organization_id` from `str | None` to `str` in `Step` in `models.py`.
>     - Change `organization_id` from `str | None` to `str` in `Task` in `schemas/tasks.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=Skyvern-AI%2Fskyvern&utm_source=github&utm_medium=referral)<sup> for d17a9d89aa253bf796f23ce755d758d8d487a68b. You can [customize](https://app.ellipsis.dev/Skyvern-AI/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/skyvern/forge/agent.py
+++ b/skyvern/forge/agent.py
@@ -210,7 +210,7 @@ class ForgeAgent:
         )
         return task, step
 
-    async def create_task(self, task_request: TaskRequest, organization_id: str | None = None) -> Task:
+    async def create_task(self, task_request: TaskRequest, organization_id: str) -> Task:
         webhook_callback_url = str(task_request.webhook_callback_url) if task_request.webhook_callback_url else None
         totp_verification_url = str(task_request.totp_verification_url) if task_request.totp_verification_url else None
         task = await app.DATABASE.create_task(
--- a/skyvern/forge/agent_functions.py
+++ b/skyvern/forge/agent_functions.py
@@ -463,7 +463,7 @@ class AgentFunction:
         return
 
     async def validate_task_execution(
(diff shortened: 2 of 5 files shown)


wintonzheng (approved):
+1

which line is breaking mypy?

context: we have `organization_id: str | None = None` across the board because we "assumed" we will have public runs that everyone can get access to. that's not really the case today and we might revisit one day.

asherf:
> +1
> 
> 
> 
> which line is breaking mypy?
> 
> 
> 
> context: we have `organization_id: str | None = None` across the board because we "assumed" we will have public runs that everyone can get access to. that's not really the case today and we might revisit one day.

The issue I noticed in my other PR is that because None is "allowed" I had to allow it in the method signature for _get_storage_class_for_org which annoyed me so I am fixing things....

PR #2623, opened 2025-06-06
Replace insecure Node.js installation with package verification

asherf:
* Remove curl-to-bash installation method due to security risks
* Implement secure installation using signed package verification
* Pin NodeSource repository public key in repository to prevent tampering
* Mitigate potential supply chain attacks by ensuring key integrity


deveshruttala-edu:

We’d love your feedback!

nodesource-repo.gpg.key
```diff
@@ -0,0 +1,29 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
LawyZheng, line 1: will this key expire in the future? do we need to keep it updated manully?

asherf: there is no expiration date.
the only case where this needs to be replaced is if the matching private key is compromised 
which is pretty unlikely 
I have used this pattern in the past ~6 years or so building containers. no issues.
and it protects against someone hijacking the website and modifying the bash script (which has happened to other vendors in the past)

LawyZheng: Thanks for sharing

PR #2635, opened 2025-06-08
replace custom HTTP client with n8n's requestWithAuthentication helper

Prakashmaheshwaran:


integrations/n8n/nodes/Skyvern/Skyvern.node.ts
```diff
@@ -1,77 +1,46 @@
...
+    };
+    return this.helpers.requestWithAuthentication.call(this, 'skyvernApi', options);
 }
 
 export class Skyvern implements INodeType {
     description: INodeTypeDescription = {
         displayName: 'Skyvern',
         name: 'skyvern',
-        icon: 'file:skyvern.png', // eslint-disable-line
+        icon: 'file:skyvern.svg',
```
LawyZheng, line 36: do we need to add this svg file in the repo?

LawyZheng (approved):

integrations/n8n/nodes/Skyvern/Skyvern.node.ts
Prakashmaheshwaran: if we referencing as file, yes, we can also use a direct url (not recommended might break which building the node locally)

LawyZheng: the current logic is using file refering mode, right?
so let's include this file in the repo.

PR #2736, opened 2025-06-18
added raw dada logs

Prakashmaheshwaran:


suchintan (approved):
nice this is lit

skyvern/config.py
```diff
@@ -39,6 +39,7 @@ class Settings(BaseSettings):
     ENV: str = "local"
     EXECUTE_ALL_STEPS: bool = True
     JSON_LOGGING: bool = False
+    LOG_RAW_API_REQUESTS: bool = False
```
suchintan: default to true

suchintan:
```suggestion
    LOG_RAW_API_REQUESTS: bool = True
```

PR #2763, opened 2025-06-21
Browser session fix 20250621

n-sviridenko:


n-sviridenko:
This is an ugly cursor-generated fix. Didn't dig deep into the business logic. But may speed up the resolution.

PR #2772, opened 2025-06-23
Enhance task handling by adding credential ID support across various components. Store credential IDs in task models, validate them during task execution, and ensure proper credential management in the workflow context. Update schemas and service methods to accommodate these changes.

Prakashmaheshwaran:


wintonzheng (commented):
task.model is for llm models. let's add a new JSON column for storing credential IDs

PR #2788, opened 2025-06-25
Improve listbox option selection handling by using get() method and adding a check for None

Prakashmaheshwaran:


wintonzheng (approved):

skyvern/webeye/actions/handler.py
```diff
@@ -3421,7 +3421,9 @@ async def click_listbox_option(
     action: actions.SelectOptionAction,
     listbox_element_id: str,
 ) -> bool:
-    listbox_element = scraped_page.id_to_element_dict[listbox_element_id]
+    listbox_element = scraped_page.id_to_element_dict.get(listbox_element_id)
+    if listbox_element is None:
```
suchintan, line 3425: nit: you can also do `listbox_element_id not in scraped_page.id_to_element_dict`

suchintan: it's probably betteer that way

LawyZheng:
nit: theoretically, it's not supposed to be NONE.
so we don't want to check the NONE from a performance perspective.

LawyZheng (approved):

PR #2882, opened 2025-07-04
Fix LLM response format error handling

Prakashmaheshwaran:


skyvern/services/task_v2_service.py
```diff
@@ -861,6 +878,23 @@ async def run_task_v2_helper(
...
+                LOG.error(
+                    "LLM completion check response is not a valid dictionary",
+                    completion_resp=completion_resp,
+                    response_type=type(completion_resp).__name__,
+                    workflow_run_id=workflow_run_id,
+                )
+                await mark_task_v2_as_failed(
+                    task_v2_id=task_v2_id,
+                    workflow_run_id=workflow_run_id,
+                    failure_reason=f"LLM completion check response format error: Expected JSON object but received {type(completion_resp).__name__}. This usually indicates the LLM model is not following the expected response format. Please try using a different model such as qwen2.5-vl, qwen2.5-14b, or gemma-3.",
```
wintonzheng, line 893: our users are not always open source users. this failure reason is very confusing to our cloud users and they won't even have the control of what model to user in cloud.

suchintan: This is probably confusing even for the open source users

PR #2975, opened 2025-07-17
Add Prompt-Based Loop Block Support to V1 Workflows

pedrodebruin:
Description
This PR introduces support for prompt-based loop blocks in the V1 workflow system, similar to the functionality available in TaskV2. The main changes include:

Backend:
- Added a loop_prompt field to the workflow YAML and block models.
- Updated the workflow service to handle prompt-based loops while preserving backward compatibility.

Frontend:
- Enhanced the LoopNode component to include a prompt input field with clear user instructions.
- Updated the UI to allow users to define loop logic using natural language prompts.

Motivation:
- This enhancement allows users to define loop logic in V1 workflows using prompts, making workflow creation more flexible and user-friendly, and bringing V1 workflows closer to the capabilities of TaskV2.

Notes:
- All changes should be backward compatible.

### DIFF
--- a/skyvern-frontend/src/routes/workflows/editor/nodes/LoopNode/LoopNode.tsx
+++ b/skyvern-frontend/src/routes/workflows/editor/nodes/LoopNode/LoopNode.tsx
@@ -36,7 +36,7 @@ function LoopNode({ id, data }: NodeProps<LoopNode>) {
   const thisBlockIsPlaying =
     urlBlockLabel !== undefined && urlBlockLabel === label;
   const [inputs, setInputs] = useState({
-    loopVariableReference: data.loopVariableReference,
+    loopValueOrPrompt: data.loopValueOrPrompt,
   });
 
   const isFirstWorkflowBlock = useIsFirstBlockInWorkflow({ id });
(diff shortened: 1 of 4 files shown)


skyvern/forge/sdk/workflow/service.py
```diff
@@ -1810,7 +1810,37 @@ async def block_yaml_to_block(
...
+                
+                # Create context parameter for the loop
+                loop_values_key = f"loop_values_{block_yaml.label}"
+                loop_over_parameter = ContextParameter(
+                    key=loop_values_key,
+                    source=extraction_output_parameter,
+                )
+                
+                # Add the context parameter to the workflow context
+                app.WORKFLOW_CONTEXT_MANAGER.add_context_parameter(
```
wintonzheng: no need to context parameter to the context manager so we can remove this

wintonzheng: this function (block_yaml_to_block) is called when we're creating a workflow.

WORKFLOW_CONTEXT_MANAGER should be used when running a workflow (workflow_run_id will be available at that point)

scripts/create_organization.py
```diff
@@ -1,3 +1,6 @@
+from dotenv import load_dotenv
+load_dotenv()
```
suchintan, line 2: huh I wonder if this was causing issues

skyvern-frontend/src/routes/workflows/editor/nodes/LoopNode/LoopNode.tsx
```diff
@@ -128,6 +137,19 @@ function LoopNode({ id, data }: NodeProps<LoopNode>) {
                 }}
               />
             </div>
+            <div style={{ marginBottom: 12 }}>
```
suchintan: What do you think about scrapping this textbox?

What if users just wrote their prompts in the original textbox and we decided what to do in the backend?

eg. If a user passed in a variable that resolved successfully, we use the existing behaviour
eg if a user passed in a prompt, use the new flow

pedrodebruin:
Closing this PR in favor of a new one that is up-to-date with the skyvern `main` and actually succeeds at the task

PR #3027, opened 2025-07-24
Natural language support for Loop Block

pedrodebruin:
# Add Natural Language Support for ForLoopBlock

This PR adds the ability to use natural language prompts in `ForLoopBlock`s, making workflows more intuitive to write while maintaining backward compatibility with existing parameter-based loops.

### 1. Natural Language Processing in ForLoopBlock
- Added `process_natural_language_prompt()` method that:
  - Takes a natural language description of what to loop over and do
  - Uses LLM to break it down into an extraction block (to get items to iterate over) and subsequent blocks (actions to perform on each item)
  - Automatically structures the extraction schema to ensure it returns an array in `results`
  - Returns both the reference to extracted data and the generated blocks

### 2. Enhanced Parameter Resolution
- Modified `try_parse_jinja_template()` to handle multiple access patterns:
  1. Direct list access (e.g., `get_top_articles_output`)
  2. Nested results access (e.g., `extraction_output.extracted_information.results`)
  3. Fallback to exact path
- Added smarter parameter path detection to preserve backward compatibility

### 3. Improved URL Navigation
- Rewrote `UrlBlock` to inherit directly from `Block` instead of `BaseTaskBlock`
- Implemented direct URL navigation using `browser_state.navigate_to_url()` rather than trying to find clickable elements
- This improves reliability when navigating to URLs extracted from previous steps

### 4. Enhanced Logging
- Added detailed execution plan logging showing:
  - Original natural language prompt
  - Inferred block types and their order
  - Extraction schema
  - Total number of blocks
- Improved debug logging throughout the parameter resolution process

## Example Usage

Before:
```yaml
- label: analyze_results
  block_type: for_loop
  loop_blocks:
    - label: goto_article
      block_type: goto_url
      url: "{{ current_value.url }}"
    - label: extract_text
      block_type: extraction
      data_extraction_goal: "Summarize the content"
  loop_variable_reference: extract_products_output.extracted_information.results
```

After (using natural language):
```yaml
- label: analyze_results
  block_type: for_loop
  loop_blocks: []  # Blocks will be generated from the prompt
  loop_variable_reference: "Extract URLs of top 3 articles and for each one: visit the URL and summarize its content"
```

## Backward Compatibility
- Existing workflows using explicit `loop_variable_reference` paths continue to work unchanged
- The system detects if a reference looks like a parameter path (contains dots or `_output`) and skips natural language processing
- Only attempts natural language processing if Jinja template parsing fails AND the reference doesn't look like a parameter path

## Testing
- Added test workflows:
  - `ycombinator_article_summarizer_test_2.json`: Tests natural language loop with URL extraction and navigation
  - `test_old_workflow.yaml`: Verifies backward compatibility with parameter-based loops

### Block Generation Process
1. LLM receives the natural language prompt and current page context
2. Generates:
   - Extraction goal and schema for getting items to iterate over
   - List of subsequent blocks with their parameters
3. System validates and creates:
   - Initial extraction block (executed immediately)
   - Loop blocks (executed for each extracted item)

### Schema Validation
- Ensures extraction blocks always return an array in `results`
- Automatically wraps raw schemas in the correct structure:
```json
{
  "type": "object",
  "properties": {
    "results": {
      "type": "array",
      "items": { ... }
    }
  },
  "required": ["results"]
}
```

### Block Type Mapping
Maintains a clear mapping of block types to their required fields:
```python
block_type_map = {
    "goto_url": (UrlBlock, {"url"}),
    "navigation": (NavigationBlock, {"navigation_goal"}),
    "extraction": (ExtractionBlock, {"data_extraction_goal"}),
    "file_download": (FileDownloadBlock, {"url"}),
    "task_v2": (TaskV2Block, {"prompt"}),
    "action": (ActionBlock, set()),
    "login": (LoginBlock, set()),
    "wait": (WaitBlock, {"wait_sec"}),
}
```

## Future Improvements
1. Add support for more complex loop patterns
2. Enhance schema inference based on the natural language description
3. Add ability to specify custom block parameters in natural language
4. Improve error messages when natural language processing fails

### DIFF
--- a/skyvern/forge/sdk/workflow/models/block.py
+++ b/skyvern/forge/sdk/workflow/models/block.py
@@ -950,8 +950,20 @@ class ForLoopBlock(Block):
         parameter_value = None
         if self.loop_variable_reference:
             LOG.debug("Processing loop variable reference", loop_variable_reference=self.loop_variable_reference)
+            
+            # Check if this looks like a parameter path (contains dots and/or _output)
+            is_likely_parameter_path = (
+                "." in self.loop_variable_reference or 
+                "_output" in self.loop_variable_reference
+            )
+            
+            # Try parsing as Jinja template
             parameter_value = self.try_parse_jinja_template(workflow_run_context)
-            if parameter_value is None:
+            
+            # Only try natural language processing if:
+            # 1. Jinja parsing failed AND
+            # 2. This doesn't look like a parameter path
+            if parameter_value is None and not is_likely_parameter_path:
                 try:
                     # If Jinja parsing failed, treat as natural language and process
                     new_ref, new_blocks = await self.process_natural_language_prompt(
@@ -993,6 +1005,10 @@ class ForLoopBlock(Block):
                 except Exception as e:
                     LOG.error("Error during natural language processing", error=str(e), exc_info=True)
                     raise
+            elif parameter_value is None and is_likely_parameter_path:
+                # If this looks like a parameter path but parsing failed, raise a more specific error
+                LOG.error("Failed to parse parameter path", loop_variable_reference=self.loop_variable_reference)
+                raise ValueError(f"Failed to parse parameter path: {self.loop_variable_reference}")
 
         if parameter_value is None:
             raise ValueError("No parameter value found")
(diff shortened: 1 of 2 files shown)


skyvern/forge/sdk/workflow/models/block.py
```diff
@@ -891,85 +893,354 @@ def get_all_parameters(
...
+        browser_state = app.BROWSER_MANAGER.get_for_workflow_run(workflow_run_id)
+        if browser_state:
+            page = await browser_state.get_working_page()
+            if page:
+                current_url = await SkyvernFrame.get_url(frame=page)
+
+        LOG.debug("Processing natural language prompt", prompt=self.loop_variable_reference, current_url=current_url)
+
+        # Generate prompt for LLM to break down the instruction
+        prompt = prompt_engine.load_prompt(
```
suchintan: This is definitely the right solution (approach) to the problem, but perhaps introduces a decent amount of new logic to our ecosystem

1. A new custom prompt (which has to be maintained / benchmarked) within the for loop ecosystem
2. A new access point for the browser (Which may or may not need to be refactored soon)

Anotehr approach to solving this problem is to have the for loop block do 2 operations:
1. Extract block (w/ a pre-defined output schema!)
2. Loop over the output of the extract block

Thsi is how our customers (unintuitively) do it today, and bundling them in this way has a few advantages:
(1) It re-uses extraction block, so any improvements to that block will automatically help here
(2) The browser code remains very separate from the workflows code (as it is today)

What do you think?

skyvern/forge/sdk/workflow/models/block.py
```diff
@@ -891,85 +893,354 @@ def get_all_parameters(
...
+                created_at=now,
+                modified_at=now,
+                parameter_type=ParameterType.OUTPUT,
+                description=f"Output parameter for extraction block {self.label}_extraction"
+            )
+        )
+        blocks.append(extraction_block)
+
+        # Map of block types to their classes and required fields
+        block_type_map = {
```
suchintan: This should not be defined in-line

skyvern/forge/sdk/workflow/models/block.py
pedrodebruin: The concerns are valid and I agree. Let's discuss the alternative

> Anotehr approach to solving this problem is to have the for loop block do 2 operations:
> 
> Extract block (w/ a pre-defined output schema!)
> Loop over the output of the extract block
> Thsi is how our customers (unintuitively) do it today, and bundling them in this way has a few advantages:
> (1) It re-uses extraction block, so any improvements to that block will automatically help here
> (2) The browser code remains very separate from the workflows code (as it is today)
> 
> What do you think?

A few comments:
- My current implementation forces the first block to be an Extraction block. The prompt in `for-loop-breakdown.j2` also states we should start with an extraction block. But I agree, letting the AI construct the schema allows for more failure
- This means we need to change the Loop Block FE UI to allow for a schema to be passed?

skyvern/forge/sdk/workflow/models/block.py
pedrodebruin: removed code

wintonzheng:
Would it still work if the `loop_over_parameter_key` is defined in the loop block? example and screenshot:

add an input parameter `list_input`:
```
      {
        "key": "list_input",
        "description": null,
        "parameter_type": "workflow",
        "workflow_parameter_type": "json",
        "default_value": [
          "1",
          "2"
        ]
      }
```

and here's the loop block definition:
```
      {
        "label": "block_1",
        "continue_on_failure": false,
        "block_type": "for_loop",
        "loop_over_parameter_key": "list_input",
        "loop_blocks": [],
        "loop_variable_reference": "{{list_input}}",
        "complete_if_empty": false
      }
```

Screenshot 2025-07-25 at 8 01 59 AM

pedrodebruin:
@wintonzheng I think so. I just tested it with the input parameter below and it seems to work. It successfully goes each page. It extracted the author of the first page but it's having a hard time extracting the author for the other page. I think it's an unrelated issue related to parsing the webpage.

list_input parameter I used:
```
[
  {
    "title": "It's DE9, Not DB9",
    "url": "https://news.sparkfun.com/14298"
  },
  {
    "title": "Who has the fastest F1 website (2021)",
    "url": "https://jakearchibald.com/2021/f1-perf-part-3/"
  }
]
```

image

PR #3105, opened 2025-08-05
fix onepassword router

LawyZheng:


skyvern/forge/sdk/routes/credentials.py
```diff
@@ -375,7 +375,7 @@ async def get_credentials(
 
 
 @base_router.get(
-    "/credentials/onepassword",
+    "/credentials/onepassword/get",
```
LawyZheng, line 378: i really don't understand why the router is working for`/credentials/onepassword` post but not working for `/credentials/onepassword` get when we set the router path as original

however, when i change the router to `/credentials/onepassword/get`, it works.

wintonzheng: It's because we have a router `/credentials/{credential_id}` which conflicts with `/credentials/onepassword`

LawyZheng: YOU'RE RIGHT!

PR #3387, opened 2025-09-07
Also use in bitwarden parameters

stenn930:


wintonzheng (approved):

stenn930:
> would be great to apply _resolve_parameter_value to other parameter rendering functions that start with `register_`

Agree. But I don't have the ability to verify/test those changes, so uncomfortable modifying that

PR #3399, opened 2025-09-10
enable code block frontend by default

wintonzheng:


LawyZheng:
should change this logic to enable it by default https://github.com/Skyvern-AI/skyvern/blob/3b8789d04517fc3a9d2bac7cc6320a3b6c07d6d1/skyvern-frontend/src/routes/workflows/editor/panels/WorkflowNodeLibraryPanel.tsx#L15-L16

skyvern-frontend/.env.example
```diff
@@ -14,4 +14,4 @@ VITE_SKYVERN_API_KEY=YOUR_API_KEY
 VITE_ENABLE_LOG_ARTIFACTS=false
 
 # Enable code block node
-VITE_ENABLE_CODE_BLOCK=false
+VITE_ENABLE_CODE_BLOCK=true
```
LawyZheng, line 17: it's an example file. if user don't copy the example file when he started the project, codeblock will be still disabled in the frontend by default

wintonzheng: https://github.com/Skyvern-AI/skyvern/blob/d09c1c865cd2a753855db9c656f81b710eafc8c0/run_ui.sh#L11

the run ui script is copy pasting the .env file if it doesn't exist

PR #3614, opened 2025-10-05
feat: self healing skyvern api key

gswangg:


skyvern/forge/sdk/services/local_org_auth_token_service.py
```diff
@@ -0,0 +1,74 @@
...
+
+LOG = structlog.get_logger()
+PROJECT_ROOT = Path(__file__).resolve().parents[4]
+ROOT_ENV_PATH = PROJECT_ROOT / ".env"
+FRONTEND_ENV_PATH = PROJECT_ROOT / "skyvern-frontend" / ".env"
+SKYVERN_LOCAL_ORG = "Skyvern-local"
+SKYVERN_LOCAL_DOMAIN = "skyvern.local"
+
+
+def _write_env(path: Path, key: str, value: str) -> None:
```
suchintan, line 19: We need to be careful merging this so we don't allow this in our cloud instance by accident!

skyvern/forge/sdk/services/local_org_auth_token_service.py
```diff
@@ -0,0 +1,74 @@
...
+from skyvern.config import settings
+from skyvern.forge import app
+from skyvern.forge.sdk.core import security
+from skyvern.forge.sdk.schemas.organizations import Organization, OrganizationAuthTokenType
+from skyvern.forge.sdk.services.org_auth_token_service import API_KEY_LIFETIME
+
+LOG = structlog.get_logger()
+PROJECT_ROOT = Path(__file__).resolve().parents[4]
+ROOT_ENV_PATH = PROJECT_ROOT / ".env"
+FRONTEND_ENV_PATH = PROJECT_ROOT / "skyvern-frontend" / ".env"
```
suchintan: will this work for pip installed skyvern?

skyvern/forge/sdk/routes/internal_auth.py
```diff
@@ -0,0 +1,126 @@
...
+        return False
+    try:
+        addr = ipaddress.ip_address(host)
+    except ValueError:
+        return False
+    return addr.is_loopback or addr.is_private
+
+
+def _require_local_access(request: Request) -> None:
+    if settings.ENV != "local":
```
suchintan, line 45: nice

skyvern/forge/sdk/routes/internal_auth.py
```diff
@@ -0,0 +1,126 @@
...
+
+import structlog
+from fastapi import APIRouter, HTTPException, Request, status
+
+from skyvern.config import settings
+from skyvern.forge import app
+from skyvern.forge.sdk.services.local_org_auth_token_service import fingerprint_token, regenerate_local_api_key
+from skyvern.forge.sdk.services.org_auth_service import resolve_org_from_api_key
+
+router = APIRouter(prefix="/internal/auth", tags=["internal"])
```
suchintan, line 13: can you hide this from localhost:8000/docs by default?

skyvern/forge/sdk/services/local_org_auth_token_service.py
gswangg: i think the file writing would, but this change relies on the frontend dev server behavior hot reloading once the env file is modified, but you'd probably be doing `skyvern run ui` or something like that when using pip installed skyvern, so that wouldn't work. i'd have to think about how to support that usecase...

gswangg: i pushed a patch that addresses this but testing it locally in the different setups and refining it is taking longer than expected. specifically, i think there might be some more refinements that need to be done to how the backend .env is getting written to in wheel-installed envs. im seeing some unexpected behavior that's slowing me down though so just working through those setup issues.

gswangg: > will this work for pip installed skyvern?

yes! now it will.

skyvern/forge/sdk/routes/internal_auth.py
gswangg: yes! now it's hidden

gswangg:
Note that this PR was rebased on #3691, but could be made independent of it pretty trivially.

Because of the rebase, the diff is a bit hard to read. Here is a diff between my two PR branches: https://github.com/gswangg/skyvern/compare/chore/env-path-refactor...gswangg:skyvern:feature/self-healing-skyvern-api-key

skyvern/cli/commands.py
```diff
@@ -54,5 +56,5 @@ def init_browser_command() -> None:
 
 
 if __name__ == "__main__":  # pragma: no cover - manual CLI invocation
-    load_dotenv()
+    load_dotenv(str(resolve_backend_env_path()))
```
wintonzheng: why not make the resolve_backend_env_path return str so we don't have to call `str(resolve_backend_env_path())`?

gswangg:
self-healing-skyvern

preview^

beginning banner layout shift is demonstrating that the banner flexes with the side/nav.

note: we use sessionStorage to persist the api key for static builds that don't hot-reload the .env.
the side effect here is that for static builds, the persisted key is local to the tab. if you open a new tab without restarting the ui (`skyvern run ui`, `npm run build&&npm run preview`, etc.), you'll get the 'missing key' banner again. restarting the ui after regeneration will always sync everything up though. the reason we do this is that using more durable storage, i.e. localStorage, persists the key in ways that could cause very confusing configuration problems for users who might be setting up skyvern from scratch often (notably devs) which would cause you to have to manually clear localStorage, which is very unintuitive and tedious.

skyvern/cli/commands.py
gswangg: fixed in #3691

skyvern-frontend/src/components/SelfHealApiKeyBanner.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+      setApiKeyHeader(apiKey);
+
+      const fingerprintSuffix = fingerprint
+        ? ` (fingerprint ${fingerprint})`
+        : "";
+
+      toast({
+        title: "API key regenerated",
+        description: `Requests now use the updated key automatically${fingerprintSuffix}.`,
```
wintonzheng: Is it possible for the backend to return the path to where the .env file and show it here as well, so that users know where this API key is stored/updated?

gswangg: Done

image

gswangg: added small note about sessionStorage to the toast too in a subsequent patch

https://github.com/gswangg/skyvern/blob/07efd72fb2d05285259bb2c09f72b18ac68e58ef/skyvern-frontend/src/components/SelfHealApiKeyBanner.tsx#L126

gswangg: added server restart text to production build banner and toast

image


image

PR #3994, opened 2025-11-13
feat: add hover action support

mohamedmamdouh22:
- Implemented HOVER action type across frontend and backend to enable mouse-over interactions
- Updated LLM prompts to recognize hover-triggered UI elements and generate appropriate hover actions
- Added hover action handling in script generation with configurable hold duration
- Fixed markup issue in CLI output to prevent Rich console formatting errors

---

🖱️ This PR introduces comprehensive hover action support to Skyvern, enabling the automation framework to handle mouse-over interactions that reveal hidden UI elements like dropdown menus, tooltips, and hover-triggered buttons. The implementation spans both frontend and backend components, including LLM prompt updates to intelligently detect and interact with hover-dependent UI elements.

<details>
<summary>🔍 <strong>Detailed Analysis</strong></summary>

### Key Changes
- **Action Type System**: Added `HOVER` as a new action type across the entire codebase, including TypeScript definitions, Python enums, and action handlers
- **Frontend Integration**: Updated React components to display hover actions with appropriate icons and labels in the task detail views
- **LLM Prompt Enhancement**: Modified AI prompts to recognize hover-triggered UI elements and generate appropriate hover actions before attempting clicks
- **DOM Analysis**: Enhanced JavaScript DOM utilities to detect hover-only elements and mark them appropriately for the automation engine
- **Script Generation**: Added hover action support in generated scripts with configurable hold duration parameters

### Technical Implementation
```mermaid
sequenceDiagram
    participant LLM as AI Agent
    participant DOM as DOM Scanner
    participant Action as Action Handler
    participant Browser as Browser/Page
    
    DOM->>DOM: Detect hover-only elements
    DOM->>LLM: Mark elements with hoverOnly flag
    LLM->>Action: Generate HOVER action
    Action->>Browser: Execute hover with optional hold_seconds
    Browser->>Browser: Reveal hidden UI elements
    Action->>LLM: Continue with next action (e.g., CLICK)
```

### Impact
- **Enhanced Automation Capability**: Enables interaction with modern web UIs that rely heavily on hover states for navigation and functionality
- **Improved Success Rates**: Reduces task failures caused by attempting to click hidden elements that only appear on hover
- **Better User Experience**: Provides visual feedback in the frontend for hover actions, making automation workflows more transparent and debuggable

</details>

_Created with [Palmier](https://www.palmier.io)_

----

> [!IMPORTANT]
> Adds hover action support across Skyvern, updating frontend, backend, and script generation to handle hover interactions.
> 
>   - **Behavior**:
>     - Adds `HOVER` action type to `ActionTypes` in `types.ts` and updates `ActionTypePill.tsx` and `ActionTypePillMinimal.tsx` to display hover actions.
>     - Enhances LLM prompts in `extract-action-static.j2` and `extract-action.j2` to recognize hover-triggered UI elements.
>     - Implements hover action handling in `generate_script.py` with configurable hold duration.
>   - **DOM and Element Handling**:
>     - Updates `domUtils.js` to detect hover-only elements and mark them with `hoverOnly` flag.
>     - Modifies `dom.py` to include `hover_to_reveal()` method in `SkyvernElement` class.
>   - **Error Handling**:
>     - Adds `FailToHover` exception in `exceptions.py`.
>   - **Misc**:
>     - Fixes CLI output markup issue in `init_command.py` to prevent Rich console formatting errors.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=Skyvern-AI%2Fskyvern&utm_source=github&utm_medium=referral)<sup> for 0fda52f780d1eab6a2becd89d87afa42c36683c5. You can [customize](https://app.ellipsis.dev/Skyvern-AI/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

## Summary by CodeRabbit

* **New Features**
  * Added Hover action support across the UI, including hover badges/icons and optional hold duration.

* **Behavior / Reliability**
  * System detects and marks hover-only elements; performs hover+wait, scroll-into-view and retries to reveal them.
  * Hover failures now surface clearer failure results; agent preserves/restores prior hover scroll position when applicable.

* **Documentation**
  * Prompts and action docs updated to include HOVER semantics and usage guidance.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


skyvern/forge/prompts/skyvern/extract-action-static.j2
```diff
@@ -3,9 +3,10 @@ Include only the elements that are relevant to the user goal, without altering o
 Accurately interpret and understand the functional significance of SVG elements based on their shapes and context within the webpage.
 Use the user details to fill in necessary values. Always satisfy required fields if the field isn't already filled in. Don't return any action for the same field, if this field is already filled in and the value is the same as the one you would have filled in.
 MAKE SURE YOU OUTPUT VALID JSON. No text before or after JSON, no trailing commas, no comments (//), no unnecessary quotes, etc.
-Each interactable element is tagged with an ID. Avoid taking action on a disabled element when there is an alternative action available.
+Each interactable element is tagged with an ID. Avoid taking action on a disabled element when there is an alternative action available. The `id` you output MUST exactly match one of the IDs from the provided elements list—never invent IDs (e.g., `INSTER_CARD_WRAPPER`). If the user refers to an element that isn't listed, choose the closest matching real element ID or explain why no suitable element exists.
```
LawyZheng: why do we need this change? have you run into any issues without the change?

i'm asking becasue:
 we've supported the `solve_captcha`, `Wait` action, and going to support the actions like`reload_page`, `open_tab` in the future. These actions do not need the ID. i think this prompt might bring a side-affect for these actions.

skyvern/forge/prompts/skyvern/extract-action-static.j2
```diff
@@ -18,7 +19,7 @@ Reply in JSON format with the following keys:
...
...TERMINATE". "CLICK" is an element you'd like to click. "HOVER" is used to move the mouse over an element (for example to reveal hover-only menus) without clicking. "INPUT_TEXT" is an element you'd like to input text into. "UPLOAD_FILE" is an element you'd like to upload a file into. "SELECT_OPTION" is an element you'd like to select an option from. "WAIT" action should be used only if there are no actions to take and there is some indication on screen that waiting could yield more actions. "WAIT" should not be used if there are actions to take such as hovering or clicking. "SOLVE_CAPTCHA" should be used if there's a captcha to solve on the screen. "COMPLETE" is used when the user goal has been achieved AND if there's any data extraction goal, you should be able to get data from the page. Never return a COMPLETE action unless the user goal is achieved. "TERMINATE" is used to terminate the whole task with a failure when it doesn't seem like the user goal can be achieved. Do not use "TERMINATE" if waiting could lead the user towards the goal. Only return "TERMINATE" if you are on a page where the user goal cannot be achieved. All other actions are ignored when "TERMINATE" is returned.
```
LawyZheng: why revert the prompt like `{{' "CLOSE_PAGE" is used to close the current page when it is impossible to achieve the user goal on the current page.' if has_magic_link_page else ''}}`

skyvern/forge/prompts/skyvern/extract-action-static.j2
```diff
@@ -3,9 +3,10 @@ Include only the elements that are relevant to the user goal, without altering o
...
-Each interactable element is tagged with an ID. Avoid taking action on a disabled element when there is an alternative action available.
+Each interactable element is tagged with an ID. Avoid taking action on a disabled element when there is an alternative action available. The `id` you output MUST exactly match one of the IDs from the provided elements list—never invent IDs (e.g., `INSTER_CARD_WRAPPER`). If the user refers to an element that isn't listed, choose the closest matching real element ID or explain why no suitable element exists.
 If you see any information in red in the page screenshot, this means a condition wasn't satisfied. prioritize actions with the red information.
 If you see a popup in the page screenshot, prioritize actions on the popup.
+If the UI hints that a control (like a CTA button) only appears after hovering a card, tile, or model name, always return a HOVER action on the element that needs hovering and explicitly wait for an on-hover UI update before attempting the next action. Use a WAIT action after the hover if the new element needs extra time to appear.
```
LawyZheng, line 9: why can't we move these description to `action_type` field?

skyvern/webeye/scraper/scraper.py
```diff
@@ -111,6 +111,9 @@ def json_to_html(element: dict, need_skyvern_attrs: bool = True) -> str:
     tag = element["tagName"]
     attributes: dict[str, Any] = copy.deepcopy(element.get("attributes", {}))
 
+    if element.get("hoverOnly"):
+        attributes["data-skyvern-hover-only"] = "true"
```
LawyZheng: can we shorten the attr name? It's wasting tokens.

skyvern/webeye/actions/handler.py
```diff
@@ -2260,6 +2298,7 @@ async def fc_func(fc: FileChooser) -> None:
     :param css: css of the element to click
     """
     try:
+        await skyvern_element.hover_to_reveal()
```
LawyZheng, line 2351: Why do we need this?

skyvern/webeye/scraper/domUtils.js
```diff
@@ -411,6 +411,40 @@ function hasASPClientControl() {
...
+      parentClass.includes("item")
+    ) {
+      // This element might be revealed on parent hover
+      return true;
+    }
+    parent = parent.parentElement;
+  }
+
+  return false;
+}
```
LawyZheng, line 450: i know it's an initial version, but what's the next plan to optimize it?


The scope, like this, might be very wide -- which means it might increase tons of tokens in the LLM call. do you have a plan to observe the change?
```
      parentClass.includes("hover") ||
      parentClass.includes("card") ||
      parentClass.includes("item")
```

LawyZheng:
appreciate the change!

Generally speaking, i think the PR is good.
Hover action is a great idea, and we should definitely support actions like this. But i just have a question:

At the beginning of the scraping/step, Skyvern will scroll to the top of the page, and scroll down to take screenshots + parse the DOM. Will the hover thing be broken when the page is scrolled up and down?

For example:
step 1 -> give an action to hover on an element, and complete the current step.
step 2 -> beginning scraping(scrolling up and down), what will happen?


It will be better if you can provide a real case that can be unblocked by adding the hover action.

cc @wintonzheng @suchintan  for another review.

suchintan:
> appreciate the change!
> 
> Generally speaking, i think the PR is good. Hover action is a great idea, and we should definitely support actions like this. But i just have a question:
> 
> At the beginning of the scraping/step, Skyvern will scroll to the top of the page, and scroll down to take screenshots + parse the DOM. Will the hover thing be broken when the page is scrolled up and down?
> 
> For example: step 1 -> give an action to hover on an element, and complete the current step. step 2 -> beginning scraping(scrolling up and down), what will happen?
> 
> It will be better if you can provide a real case that can be unblocked by adding the hover action.
> 
> cc @wintonzheng @suchintan for another review.

The context is here!
https://skyvern.slack.com/archives/C09R66J9L4X/p1762792387368639

skyvern/webeye/actions/handler.py
mohamedmamdouh22: hover_to_reveal() simulates moving the mouse over the element before attempting to click or follow its link. Many menus/buttons are hidden  until hovered, so clicking them directly fails. We mark some nodes as “hover-only” during scraping, and when an action targets one, hover_to_reveal() expands it so the subsequent navigate_to_a_href/locator.click has something visible and interactable to hit.

skyvern/webeye/scraper/scraper.py
mohamedmamdouh22: okay done

skyvern/forge/prompts/skyvern/extract-action-static.j2
mohamedmamdouh22: yeah actually without this instruction the model was hovering on the wrong element, to address the future issue you mentioned, what do you think about this change?
```suggestion
Each interactable element is tagged with an ID. Avoid taking action on a disabled element when there is an alternative action available. For element-based actions (CLICK, HOVER, INPUT_TEXT, UPLOAD_FILE, SELECT_OPTION), the `id` you output MUST exactly match one of the IDs from the provided elements list—never invent IDs. If the user refers to an element that isn't listed, choose the closest matching real element ID or explain why no suitable element exists. Actions that do not operate on a specific element (e.g., WAIT, SOLVE_CAPTCHA, RELOAD_PAGE, OPEN_TAB) should not provide an `id`.
```

skyvern/forge/prompts/skyvern/extract-action-static.j2
mohamedmamdouh22: I’d keep it outside the action_type field because it’s a cross-cutting rule, not specific to any single action. The action list is already dense, and burying the “don’t invent IDs / non-element actions omit IDs” guidance there makes it easy to miss. Having it up front keeps the guardrail obvious, while the id field description still references the same expectation. If we need to reinforce it further, we can add a short pointer in the id field (e.g., “Apply the earlier ID guidance before filling this”), but the standalone paragraph avoids duplication and stays more discoverable.
what do you think?

skyvern/forge/prompts/skyvern/extract-action-static.j2
mohamedmamdouh22: You’re right—the version without the `{{ ... if has_magic_link_page ... }}` block was my change, and it was incorrect because it always advertises `CLOSE_PAGE` even when that action isn’t available. I’ve reverted back to the templated form so the prompt only mentions `CLOSE_PAGE` when has_magic_link_page is true. Thanks for catching it it seems like i missed it when adding the hover action instructions.

skyvern-frontend/src/routes/tasks/detail/ActionTypePill.tsx
```diff
@@ -7,6 +7,7 @@ type Props = {
 
 const icons: Partial<Record<ActionType, React.ReactNode>> = {
   click: <CursorArrowIcon className="h-4 w-4" />,
+  hover: <CursorArrowIcon className="h-4 w-4" />,
```
wintonzheng, line 10: Could you pick another icon that makes sense from https://www.radix-ui.com/icons?

skyvern-frontend/src/routes/tasks/detail/ActionTypePillMinimal.tsx
```diff
@@ -13,6 +13,7 @@ type Props = {
 
 const icons: Partial<Record<ActionType, React.ReactNode>> = {
   click: <CursorArrowIcon className="h-4 w-4" />,
+  hover: <CursorArrowIcon className="h-4 w-4" />,
```
wintonzheng, line 16: same: please pick another icon that makes sense from https://www.radix-ui.com/icons?

skyvern/webeye/scraper/domUtils.js
wintonzheng: this is a great question. 

@mohamedmamdouh22 why do you think of using "card" and "item" besides "hover"?

skyvern-frontend/src/routes/tasks/detail/ActionTypePill.tsx
mohamedmamdouh22: I cant see one which is more suitable from the radix icons, please let me know if you see any one makes more sense

skyvern/webeye/scraper/domUtils.js
mohamedmamdouh22: @wintonzheng 
The extra checks for "card" and "item" in isHoverOnlyElement are a heuristic. Many modern UIs place hover-revealed controls inside card or list-item wrappers whose own hover state controls the child’s visibility. By treating ancestors with those class names as possible hover containers, the scraper marks those children as “hover-only” so they remain interactable instead of being discarded as invisible when their own bounding box is 0×0 until the parent is hovered.

mohamedmamdouh22: @LawyZheng 
Good call. Right now we walk all the way to body, which is broad and can pull in false positives and extra DOM in the prompt. Short term, we can cap the ancestor scan to ~4–5 levels to catch common card/list wrappers without fanning out to the whole page. Next steps I’d take:

- Add a depth limit to the parent walk (e.g., stop after 4–5 hops).

- Add a small metric/flag in the scraper to log how many nodes are tagged hover-only and any extra HTML we include, so we can see whether the cap actually trims token usage.

- Keep the class-name heuristic for now but watch precision: we could tighten matches (e.g., -card, card-, list-item, menu-item) and avoid generic substrings like “item” alone if it proves too noisy.

My view: a shallow cap plus slightly more specific class checks should reduce noise and prompt size without missing typical hover-only quick-actions on cards/list rows. If that’s not enough, we can switch to CSS hover rule detection or sampling actual hover states, but the depth limit is the quickest win.

skyvern/webeye/scraper/scraper.py
```diff
@@ -111,6 +111,9 @@ def json_to_html(element: dict, need_skyvern_attrs: bool = True) -> str:
     tag = element["tagName"]
     attributes: dict[str, Any] = copy.deepcopy(element.get("attributes", {}))
 
+    if element.get("hoverOnly"):
+        attributes["data-hover"] = "1"
```
LawyZheng: i'm not sure if LLM can understand what it means in HTML `<div data-hover="1">` under the instruction `"HOVER" is used to move the mouse over an element (for example, to reveal hover-only menus or buttons before clicking)`

maybe `hoverable=true` is better

skyvern/forge/prompts/skyvern/extract-action-static.j2
LawyZheng: `yeah actually without this instruction the model was hovering on the wrong element`

maybe LLM doesn't understand the `HTML + description of Hover action`, so he just gave the wrong id, wdyt?

skyvern/webeye/scraper/domUtils.js
LawyZheng: can we at least implement the depth limit in the first version?

mohamedmamdouh22: done!

skyvern/forge/prompts/skyvern/extract-action-static.j2
mohamedmamdouh22: not really sure tbh

LawyZheng:
Thanks for the contribution

mohamedmamdouh22:
[heart]         ***@***.*** reacted to your message:
________________________________

PR #4396, opened 2026-01-05
[Cache] Fix: LocalCache.set() to respect expiration parameter

CodeDotJS:
- Replace TTLCache with custom implementation that tracks per-key expiration
- Cache entries now expire at their specified times (1 hour, 1 week, etc.)
- Invalid shapes will expire correctly instead of staying cached for 4 weeks
- Add unit tests to verify expiration behavior
- Use `asyncio.Lock` for task-safe access within event loop

----

> [!IMPORTANT]
> Replace `TTLCache` with custom implementation in `LocalCache` to handle per-key expiration and add unit tests for verification.
> 
>   - **Behavior**:
>     - Replace `TTLCache` with custom implementation in `LocalCache` to track per-key expiration.
>     - Cache entries now expire at specified times (e.g., 1 hour, 1 week).
>     - Invalid shapes expire correctly instead of staying cached for 4 weeks.
>     - Use `asyncio.Lock` for thread-safe access in `LocalCache`.
>   - **Testing**:
>     - Add unit tests in `test_local_cache.py` to verify expiration behavior for custom and default expiration times.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=Skyvern-AI%2Fskyvern&utm_source=github&utm_medium=referral)<sup> for 05a98e4fb6663185d881d885dc876e7637302522. You can [customize](https://app.ellipsis.dev/Skyvern-AI/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

## Summary by CodeRabbit

* **Bug Fixes & Improvements**
  * Switched to explicit per-item expirations with async-safe access, opportunistic cleanup and periodic cleanup triggered by insertions
  * Enforced max-size eviction of oldest entries to prevent unbounded growth and improved concurrent read behavior

* **Tests**
  * Added comprehensive unit tests covering custom/default expirations, concurrent reads, insertion-triggered cleanup, and max-size enforcement

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/skyvern/forge/sdk/cache/local.py
+++ b/skyvern/forge/sdk/cache/local.py
@@ -9,8 +9,9 @@ class LocalCache(BaseCache):
     def __init__(self) -> None:
         # Use a regular dict to store (value, expiration_timestamp) tuples
         self._cache: dict[str, tuple[Any, float]] = {}
-        self._lock = asyncio.Lock()  # Async lock for task-safe access within event loop
+        self._lock = asyncio.Lock()  # Async lock for write operations
         self._default_ttl_seconds = CACHE_EXPIRE_TIME.total_seconds()
+        self._insertion_count = 0  # Track insertions for cleanup scheduling
 
     def _normalize_expiration(self, ex: Union[int, timedelta, None]) -> float:
         """Convert expiration parameter to Unix timestamp."""
@@ -36,18 +37,25 @@ class LocalCache(BaseCache):
             self._cache.pop(key, None)
 
     async def get(self, key: str) -> Any:
-        async with self._lock:
-            if key not in self._cache:
-                return None
-
+        # Try lock-free read first (dict reads are atomic in CPython)
+        # This allows concurrent reads without blocking
+        try:
             value, expiration = self._cache[key]
+        except KeyError:
+            # Key was deleted between check and access, or never existed
+            return None
 
-            # Check if expired
-            if self._is_expired(expiration):
-                del self._cache[key]
+        # Check if expired - only acquire lock if we need to delete
+        if self._is_expired(expiration):
+            async with self._lock:
+                # Double-check after acquiring lock (key might have been deleted or updated)
+                if key in self._cache:
+                    cached_value, cached_expiration = self._cache[key]
+                    if self._is_expired(cached_expiration):
+                        del self._cache[key]
                 return None
 
-            return value
+        return value
 
     async def set(self, key: str, value: Any, ex: Union[int, timedelta, None] = CACHE_EXPIRE_TIME) -> None:
         expiration_timestamp = self._normalize_expiration(ex)
@@ -61,7 +69,8 @@ class LocalCache(BaseCache):
                     del self._cache[oldest_key]
 
             self._cache[key] = (value, expiration_timestamp)
+            self._insertion_count += 1
 
-            # Opportunistic cleanup every 100 insertions
-            if len(self._cache) % 100 == 0:
+            # Opportunistic cleanup every 100 insertions (works for any MAX_CACHE_ITEM value)
+            if self._insertion_count % 100 == 0:
                 self._cleanup_expired()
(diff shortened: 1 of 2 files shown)


skyvern/forge/sdk/cache/local.py
```diff
@@ -1,20 +1,67 @@
...
-
 from skyvern.forge.sdk.cache.base import CACHE_EXPIRE_TIME, MAX_CACHE_ITEM, BaseCache
 
 
 class LocalCache(BaseCache):
     def __init__(self) -> None:
-        self.cache: TTLCache = TTLCache(maxsize=MAX_CACHE_ITEM, ttl=CACHE_EXPIRE_TIME.total_seconds())
+        # Use a regular dict to store (value, expiration_timestamp) tuples
+        self._cache: dict[str, tuple[Any, float]] = {}
+        self._lock = asyncio.Lock()  # Async lock for task-safe access within event loop
```
LawyZheng: the gloable lock for the cache will make the concurrency efficiency too low

skyvern/forge/sdk/cache/local.py
```diff
@@ -1,20 +1,67 @@
...
+        expired_keys = [key for key, (_, exp) in self._cache.items() if now > exp]
+        for key in expired_keys:
+            self._cache.pop(key, None)
 
     async def get(self, key: str) -> Any:
-        if key not in self.cache:
-            return None
-        value = self.cache[key]
-        return value
+        async with self._lock:
```
LawyZheng: why do we need to acquire the lock for all get options?

skyvern/forge/sdk/cache/local.py
```diff
@@ -1,20 +1,67 @@
...
+                del self._cache[key]
+                return None
+
+            return value
 
     async def set(self, key: str, value: Any, ex: Union[int, timedelta, None] = CACHE_EXPIRE_TIME) -> None:
-        self.cache[key] = value
+        expiration_timestamp = self._normalize_expiration(ex)
+
+        async with self._lock:
```
LawyZheng, line 63: do you have a better solution? The current solution is inefficient,
1. alll set options need to acquire the lock first
2. when the entry in the cache is over 1000, all set option will need to acquire lock + delete the older key
3. if we make `MAX_CACHE_ITEM to be configuration, and it's not % 100 == 0`(example 150), the cleanup will never be trigger when the items reached the max

CodeDotJS:
Thanks for the feedback @LawyZheng . I've made some changes to address the concurrency and cleanup issues. Could you review whether this approach works?

#### Lock efficiency concern
I've implemented lock-free reads for `get()` operations. Since dict reads are atomic in CPython, I'm doing a direct read without acquiring the lock first and only acquiring the lock when we need to delete an expired entry. I'm also using a double-check locking pattern when deleting expired entries.

The idea is that most reads (non-expired entries) can proceed in parallel without blocking.

#### "Why lock all get operations?"
I've removed the lock from the fast path entirely. Now `get()` only acquires the lock when it needs to delete an expired entry (which is a write operation). For normal reads, it's completely lock-free.

#### Cleanup and set operations
I've made two changes here:

- **Cleanup trigger:** Instead of using `len(self._cache) % 100 == 0`, I'm now using a separate `_insertion_count` counter that tracks total insertions. This way, cleanup triggers every 100 insertions regardless of what `MAX_CACHE_ITEM` is set to. So even if `MAX_CACHE_ITEM` is 150, cleanup will still trigger at insertions 100, 200, 300, etc.  

- **Set operations:** I understand the lock is still needed for `set()` operations (to atomically enforce `MAX_CACHE_ITEM`, remove oldest entries, and update counters). However, with the lock-free reads optimization, the overall performance should be much better for read-heavy workloads.  
  Do you think this is an acceptable trade-off, or would you prefer a different approach for the `set()` operations?

I've also added `KeyError` handling to make the reads more robust against race conditions.

Lemme know your thoughts on this!

PR #4636, opened 2026-02-05
Move npm build to Docker image build time

wintonzheng:
## Summary
- Move `npm run build` from container startup to Docker image build time
- Inject API key at runtime via placeholder replacement instead of rebuilding
- Reduces CPU/memory requirements for running the UI container

## Problem
The UI container was running `npm run build` at startup via the entrypoint script, which caused:
- OOM errors on 1 vCPU / 2GB memory ECS tasks
- CPU spikes during container startup
- Slow container startup times

## Solution
- Run `npm run build` during `docker build` (image creation)
- Use a placeholder (`__SKYVERN_API_KEY_PLACEHOLDER__`) for the API key
- At container startup, use `sed` to inject the actual API key into pre-built JS files
- Start servers directly without rebuilding

## Test plan
- [ ] Build the Docker image: `docker build -f Dockerfile.ui -t skyvern-ui .`
- [ ] Run the container and verify it starts quickly without CPU/memory spikes
- [ ] Verify the UI loads correctly in browser
- [ ] Verify API key injection works when `.streamlit/secrets.toml` is present


entrypoint-skyvernui.sh
```diff
@@ -2,9 +2,15 @@
...
+# Extract API key from secrets file
+VITE_SKYVERN_API_KEY=$(sed -n 's/.*cred\s*=\s*"\([^"]*\)".*/\1/p' .streamlit/secrets.toml 2>/dev/null || echo "")
+
+# Inject API key into pre-built JS files (replace placeholder)
+if [ -n "$VITE_SKYVERN_API_KEY" ]; then
+    find /app/dist -name "*.js" -exec sed -i "s/__SKYVERN_API_KEY_PLACEHOLDER__/$VITE_SKYVERN_API_KEY/g" {} \;
+fi
+
+# Start the servers (no rebuild needed)
+node localServer.js & node artifactServer.js & wait
```
noqcks: why did we replace npm start? were we not running two processes before?

imo should use something like tini instead of launching two processses + wait, becuase tini will give us SIGTERM signal forwarding, restarting child processes, etc

wintonzheng:
Thanks for the feedback! Added tini as an init process for proper signal handling and zombie reaping.

**To answer your question about replacing `npm start`:** The original `npm start` runs `npm run serve & npm run run-artifact-server`, where `npm run serve` does `npm run build && node localServer.js`. So yes, it was already running two processes, BUT it was also running the build on every container startup. The new approach pre-builds at image time (`RUN npm run build`) and only runs the two Node servers at runtime - no rebuild needed, hence the lower CPU/memory requirements.

Changes in 0dcbb9b:
- Added tini installation to Dockerfile
- Set tini as ENTRYPOINT for SIGTERM forwarding and zombie reaping
- Kept the two-process approach (localServer.js + artifactServer.js) since that's what the app needs

Issue #4296, opened 2025-12-14
Live Streaming fails


Yasser-shrief:
The live streaming functionality (viewing the task execution in real-time) is currently broken.

<img width="1852" height="620" alt="Image" src="https://github.com/user-attachments/assets/74ee4974-5acb-42a0-9425-140e66e47e2f" />

Yasser-shrief:
Thanks for the detailed troubleshooting steps. I’ve gone through each of them (service configuration, restarting, browser setup, local environment variables, and checking logs), but unfortunately the live streaming is still not working on my end.

Could you please advise on the next steps or escalate this issue for further diagnosis? I can provide my API key and task details if needed.  @skyvern

joseluisll:
Hi, the live streaming of the workflow does not work out of the box. I have a docker-desktop setup created using LMStudio for the LLM part. Create a workflow using the welcome page to introduce the prompt, and use the Skyvern 2.0 with Code engine, I see in the logfile the following exception: 

```
[__main__] task_id=None workflow_run_id=wr_494140052738338584 organization_id=o_494139673975597314 error_type=builtins.RuntimeError error_category=ERROR exception_hash=e46b51b703a7fb04

Traceback (most recent call last):

  File "/app/run_streaming.py", line 35, in run

    workflow_run = await app.DATABASE.get_workflow_run(workflow_run_id=workflow_run_id)

                         ^^^^^^^^^^^^

  File "/app/skyvern/forge/__init__.py", line 20, in __getattr__

    raise RuntimeError("ForgeApp is not initialized. Call start_forge_app() before accessing app properties.")

RuntimeError: ForgeApp is not initialized. Call start_forge_app() before accessing app properties.
```

If I check the run_streaming.py module, it imports app from skyvern.forge, but nobody calls  start_forge_app(), as the log error states. This issue is a bug and should be easily fixed. @skyvern Look at the log that I post from the skyvern:latest docker image running in docker-desktop.

joseluisll:
With the proposed code, it works, both the screen capture and the live streaming.

PR #4393, opened 2026-01-04
feat: Full multi-session VNC support (Live view, Browser Sessions & T…

pengu-fr:
## Summary

This PR enables complete VNC streaming support for local/self-hosted deployments with multiple concurrent browser sessions. Users can now watch live browser activity, manage persistent browser sessions, and take manual control of any running workflow — all working seamlessly in parallel.

### Live View
Each workflow run gets its own dedicated VNC stream with dynamic display allocation (`:100`, `:101`, ...). Displays and ports are automatically reused when sessions end.

###️ Browser Sessions
Full integration with persistent browser sessions. VNC ports and display numbers are tracked per-session in the database, enabling reliable reconnection and session management.

### Take Control
Users can take manual control of the browser at any time. The interactor state (`agent`/`user`) now persists across VNC reconnections.

## Technical Implementation

- New `VncManager` class for per-session Xvfb, x11vnc, and websockify management
- New database columns: `display_number`, `vnc_port`, `interactor`
- Per-browser DISPLAY environment via Playwright's `env` parameter
- `--disable-gpu` flag for x11vnc compatibility
- Automatic resource reuse when sessions terminate

## Files Changed

| File | Changes |
|------|---------|
| `skyvern/webeye/vnc_manager.py` | **New** - VNC process management |
| `skyvern/forge/sdk/db/models.py` | Added `display_number`, `vnc_port`, `interactor` |
| `skyvern/forge/sdk/schemas/persistent_browser_sessions.py` | Added new fields |
| `skyvern/forge/sdk/db/agent_db.py` | Added parameters to update function |
| `skyvern/forge/sdk/routes/streaming/channels/vnc.py` | Interactor persistence, local VNC URL |
| `skyvern/forge/sdk/routes/streaming/channels/message.py` | Async interactor updates |
| `skyvern/webeye/browser_factory.py` | Per-process DISPLAY, `--disable-gpu` |
| `alembic/versions/...` | 2 new migrations |

Fixes #4392, #4296, #2968, #2779, #1930

---

🖥️ This PR implements comprehensive multi-session VNC support for local/self-hosted Skyvern deployments, enabling live browser viewing, persistent session management, and manual control across multiple concurrent workflows. It introduces a new VncManager class that dynamically allocates displays and ports per session, with database persistence for reconnection support.

<details>
<summary>🔍 <strong>Detailed Analysis</strong></summary>

### Key Changes
[...]

PR #4681, opened 2026-02-10
Integrate streaming as a Skyvern Service only in LINUX or WSL platforms

joseluisll:
### Goal

The goal of this PR is to activate the Streaming functionality **without changing the behavior or the structure** of the **original code**.

This means that he following known limitations are:

- Compatible only with **linux** or docker-environments including **WSL**.
- The original code only manages 1 streaming session. All persistent browser sessions will share the same streaming session. This means that only 1 workflow_run with a persistent browser session should be run at the same time.
- The streaming infraestructure is created only when 1 workflow_run is created with a persistent browser session requirement. When no PBS Workflow_runs have been created, and you go to the PBS list and activate a PBS, the screen is black. If you activate a WR, then all PBS screens will show the live WR screen.

### Observations

This PR solves issue #4296. The streaming service could not startup due to broken app context and missing packages into the docker image. Also, the entrypoint did not start the streaming servicies.

Here is the trace that appeared in the log:

```
[__main__] task_id=None workflow_run_id=wr_494140052738338584 organization_id=o_494139673975597314 error_type=builtins.RuntimeError error_category=ERROR exception_hash=e46b51b703a7fb04

Traceback (most recent call last):

  File "/app/run_streaming.py", line 35, in run

    workflow_run = await app.DATABASE.get_workflow_run(workflow_run_id=workflow_run_id)

                         ^^^^^^^^^^^^

  File "/app/skyvern/forge/__init__.py", line 20, in __getattr__

    raise RuntimeError("ForgeApp is not initialized. Call start_forge_app() before accessing app properties.")

RuntimeError: ForgeApp is not initialized. Call start_forge_app() before accessing app properties.
```

## Changes
- Integrate the run_streaming.py script as an internal forge_app service called StreamingService.
- Modify Dockerfile to add needed packages: x11vnc, websockify
- Modify Entrypoint-skyvern.sh: stop invoking run_streaming.py and call the "scripts/start_streaming.sh"

## Future work

I would like to make future changes to improve the integration of the StreamingService into the App.
- Integrate the start_streaming.sh script inside the StreamingService.
- Eliminate the state_file.json . The StreamingService shall read the organizations, workflow_runs and persistentbrowsersessions from the existing App Services.
- Other thoughts, to be explored.


joseluisll:
Screenshot of the applied fixed. The streaming now works!. 

image

@wintonzheng can you please review this minimal fix?. 

Thank you.

joseluisll:
I have solved also why the live streaming was not working:
- the Dockerfile for the backend does not have the x11vnc and websockify packages installed.
- the entrypoint for the backend does not call the script start_vnc_streaming.sh script, that starts XFVB, VNC and WEBSOCKIFY. 

All in all, I check that the tests pass, and also that both the screencapture and the live streaming works using the improved docker image.

Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN pip install --upgrade pip setuptools wheel
 RUN pip install --no-cache-dir --upgrade -r requirements.txt
 RUN playwright install-deps
 RUN playwright install
-RUN apt-get install -y xauth x11-apps netpbm gpg ca-certificates && apt-get clean
+RUN apt-get install -y xauth x11-apps netpbm gpg ca-certificates x11vnc websockify && apt-get clean
```
AronPerez: 🔥 Going to link a related PR [here](https://github.com/Skyvern-AI/skyvern/pull/4393/changes#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557)

entrypoint-skyvern.sh
```diff
@@ -58,5 +58,7 @@ xvfb=$!
 DISPLAY=:99 xterm 2>/dev/null &
 python run_streaming.py > /dev/null &
 
+echo "Starting live streaming services..."
+sh ./scripts/start_vnc_streaming.sh
```
AronPerez: Can we run this directly with a `./` call, or do we need shell invocation?

run_streaming.py
```diff
@@ -16,6 +17,7 @@
 
 async def run() -> None:
     await initialize_skyvern_state_file(task_id=None, workflow_run_id=None, organization_id=None)
+    start_forge_app()
```
AronPerez: Why this for streaming? Does it get called anywhere else?

AronPerez (commented):
Looks really good, like this a lot. Have some questions

Dockerfile
joseluisll: Thankyou for that insight. The problem and that PR is very big, and indeed that PR included parts of my solution as well. However, my goal is not solving that scope of the PR, which is multisession concurrent management browsers and such, my goal was to have it working for a local deployment of one user - one org. May be a feature-gate to limit the activation of this feature?

joseluisll:
Up to now, I have integrated the streaming service inside the forge_app, and used the start_event to start the service loop. It works, but with the limitations of the screencapture for 1 xfvb display 99. This is the same level of functionality, but now the run_streaming.py is obsolete and not used at all.

run_streaming.py
joseluisll: No, it is a first step to restore the functionality. The run_streaming.py was using the app class, not the instance of app that is creater LATER when the entrypoint-skyvern.sh launches the app itself. So with that, a new app is created for the streaming part. This works because the settings, db and state file is shared by files.... Next thing I worked out in the following commits is to integrate the run_streaming.py as a service in the same app and process as the skyvern forge app, as it should be.

skyvern/services/streaming/service.py
```diff
@@ -0,0 +1,200 @@
...
+
+        Args:
+            file_path: Path to save the screenshot PNG file.
+
+        Returns:
+            True if capture succeeded, False otherwise.
+        """
+        try:
+            subprocess.run(
+                f"xwd -root | xwdtopnm 2>/dev/null | pnmtopng > {file_path}",
```
wintonzheng, line 189: Did you only test in docker environment?

my concern is that this is not going to work for non linux environment

can we add an environment/os check to make sure streaming will only be enabled when these commands (xwd, pnmtopng) are legit to run?

wintonzheng:
we should have a fix for the forge app start call in streaming script: https://github.com/Skyvern-AI/skyvern/pull/4703

it looks like this PR is also introducing another functionality here to auto stream screenshot. can we make an update for the title and description of this pr?

suchintan:
@joseluisll nice work

skyvern/services/streaming/service.py
joseluisll: yes, only docker environment. I windows will not work.

joseluisll:
> we should have a fix for the forge app start call in streaming script: #4703
> 
> it looks like this PR is also introducing another functionality here to auto stream screenshot. can we make an update for the title and description of this pr?

The snapshot functionality was already there in the run_streaming.py, the screencapture for collecting images to the workflow steps. I just integrated it into the StreamingService.

skyvern/services/streaming/service.py
joseluisll: Implemented the feature gate reusing a os_detect() function in skyvern/utils.

joseluisll:
@suchintan Please review the PR if possible, to approve?

skyvern/services/streaming/service.py
```diff
@@ -0,0 +1,200 @@
...
+        Capture a screenshot using xwd/xwdtopnm/pnmtopng pipeline.
+
+        Args:
+            file_path: Path to save the screenshot PNG file.
+
+        Returns:
+            True if capture succeeded, False otherwise.
+        """
+        try:
+            subprocess.run(
```
AronPerez: `subprocess.run` is `synchronous`. We'd block the entire event loop for the duration of each screenshot capture. It's basically running every second.

Maybe wrap this in `asyncio.create_subprocess_shell()` or `asyncio.to_thread()`

entrypoint-skyvern.sh
```diff
@@ -55,8 +55,10 @@ export DISPLAY=:99
 Xvfb :99 -screen 0 1920x1080x16 &
```
AronPerez: Xvfb is started twice with different color depths. `scripts/start_vnc_streaming.sh` starts in 24-bit color, and we start in 16-bit here. Any reason we need both, or can't just rely on `./scripts/start_vnc_streaming.sh` directly?

skyvern/services/streaming/service.py
AronPerez, line 1: In run_streaming.py we have an `await initialize_skyvern_state_file(task_id=None, workflow_run_id=None, organization_id=None)` check to ensure the state file exists at startup. Do we need that here too?

skyvern/forge/forge_app.py
```diff
@@ -76,6 +82,7 @@ class ForgeApp:
     AGENT_FUNCTION: AgentFunction
     PERSISTENT_SESSIONS_MANAGER: PersistentSessionsManager
     BROWSER_SESSION_RECORDING_SERVICE: BrowserSessionRecordingService
+    STREAMING_SERVICE: StreamingService | None
```
AronPerez, line 87: 🔥

skyvern/services/streaming/service.py
joseluisll: The point of this PR is incorporate run_streaming.py as a STREAMING_SERVICE. The run_streaming.py was an external script that created an app object. I make this proposal to integrate it as an internal service, following the same pattern of the rest of services, so it is integrated in the forge_app like the rest of services. 

This PR makes the run_streaming.py obsolete.

wintonzheng:
Hey @joseluisll, thanks for putting this together and for the careful write-up of the `ForgeApp is not initialized` failure mode — really helpful context.

We're going to close this one out in favor of two PRs Andrew shipped that cover the OSS streaming story end-to-end:

- **#4891 — `feat: add browser session support to OSS Docker deployment`** (merged 2026-03-04). This is the direct replacement for what this PR was attempting: adds `x11vnc` + `websockify` to the Docker image, wires up the full `Xvfb :99 → x11vnc :5900 → websockify :6080` pipeline in the entrypoint, exposes port 6080 in `docker-compose`, and makes `create_session()` actually launch real Chromium with a dynamic CDP port (`9223–9322`) instead of just inserting a DB row. It also addresses the "only one streaming session" limitation you called out — sessions get their own browser process and CDP port rather than sharing a single VNC pipeline.
- **#4904 — `Add CDP screencast streaming and interactive input for local mode`** (merged 2026-03-24). Complements the Docker path: in local mode the frontend falls back to CDP `Page.startScreencast` over a WebSocket (no VNC server needed) and forwards clicks/keys/scroll via CDP `Input` with explicit "take control" / "cede control" gating.

Between those two, the streaming infrastructure is now wired up for both Docker and local OSS, with per-session streams instead of a shared one — so the integration in this PR isn't needed. Closing, but really appreciate you tracing this through. 🙏

joseluisll:
Great, I just wanted to help, but have limited experience.

PR #4686, opened 2026-02-10
Add GPT 5 Mini to model selector and hide GPT 5

suchintan:
## Summary
- Add **GPT 5 Mini** (`azure/gpt-5-mini`) to the model dropdown in the UI, with Gemini 2.5 Flash fallback in cloud environments
- **Hide GPT 5** from the model dropdown (marked as deprecated) while keeping backwards compatibility for users who already have it selected

## Test plan
- [ ] Verify GPT 5 Mini appears in the model selector dropdown
- [ ] Verify GPT 5 is hidden from the dropdown for new selections
- [ ] Verify existing workflows using GPT 5 still show it as "GPT 5 (deprecated)"
- [ ] Verify GPT 5 Mini uses the router fallback config in cloud

### DIFF
--- a/skyvern/forge/sdk/db/agent_db.py
+++ b/skyvern/forge/sdk/db/agent_db.py
@@ -3,7 +3,24 @@ from datetime import datetime, timedelta
 from typing import Any, List, Literal, Sequence, overload
 
 import structlog
-from sqlalchemy import Text, and_, asc, case, delete, distinct, exists, func, or_, pool, select, tuple_, update
+from sqlalchemy import (
+    Text,
+    and_,
+    asc,
+    case,
+    cast,
+    delete,
+    distinct,
+    exists,
+    func,
+    literal,
+    or_,
+    pool,
+    select,
+    tuple_,
+    update,
+)
+from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.exc import (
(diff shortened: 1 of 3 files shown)


skyvern-frontend/src/components/ModelSelector.tsx
```diff
@@ -28,6 +28,7 @@ const constants = {
 const deprecatedModelNames = new Set<string>([
   "gemini-2.5-flash-lite",
   "azure/gpt-4.1",
+  "azure/gpt-5",
```
AronPerez, line 31: 👍

skyvern/config.py
```diff
@@ -458,6 +458,7 @@ def get_model_name_to_llm_key(self) -> dict[str, dict[str, str]]:
                 },
                 "azure/gpt-4.1": {"llm_key": "AZURE_OPENAI_GPT4_1", "label": "GPT 4.1"},
                 "azure/gpt-5": {"llm_key": "AZURE_OPENAI_GPT5", "label": "GPT 5"},
```
AronPerez, line 460: We keep them in the config in-case others want to bring their own tool?

suchintan: yeah we have this in our roadmap: 

* Let people define their own LLMs under Settings

PR #4704, opened 2026-02-11
Add MCP tool analytics

marcmuon:
## Summary

- Integrate PostHog analytics into `make_result()` — the single function every MCP tool already calls — instead of manually adding `analytics.capture()` to each tool function
- Enrich `analytics.capture()` to always include system metadata (OS, skyvern version, python version, environment)
- Every `mcp_tool_call` event now includes: tool name, success/failure, timing, browser mode, session ID, plus global metadata

## Why this approach

The previous implementation added manual `analytics.capture()` calls to every MCP tool function (235+ lines across 4 files). This approach:
- **2 files, 14 lines** vs 4 files, 235 lines
- **Zero changes** to any tool file (browser.py, workflow.py, session.py, blocks.py)
- **Automatically covers** every future tool — no risk of forgetting
- **Single event name** (`mcp_tool_call`) with tool name as a property — easier to query in PostHog

## Event shape

```
event: "mcp_tool_call"
properties:
  tool: "skyvern_click"
  ok: true
  timing_ms: 1234
  error_code: null
  browser_mode: "cloud_session" | "local" | "cdp" | null
  session_id: "pbs_123" | null
  os: "darwin"
  oss_version: "0.5.2"
  machine: "arm64"
  environment: "local"
```

## Test plan

- [x] `py_compile` passes on both modified files
- [x] All pre-commit hooks pass
- [ ] Verify events appear in PostHog after MCP tool usage

### DIFF
--- a/skyvern/analytics.py
+++ b/skyvern/analytics.py
@@ -44,7 +44,7 @@ def capture(
 
     distinct_id = settings.ANALYTICS_ID
 
-    payload: dict[str, Any] = data or {}
+    payload: dict[str, Any] = {**analytics_metadata(), **(data or {})}
     try:
         posthog.capture(distinct_id=distinct_id, event=event, properties=payload)
     except Exception as e:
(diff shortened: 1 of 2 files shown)


marcmuon:
Closing — this should flow through skyvern-cloud auto-sync instead of being pushed directly to OSS.

PR #4716, opened 2026-02-12
Browser profile component

jseres:
This is a merge request which will introduce browser profiles into the UI so that users do not have to continuously login. For more detailed descriptions of specific parts of this feature, refer to commit messages.

### Remaining Concerns

- Error handling for testing credentials: when a user opts in to test + save a browser profile, when a test fails, what do we do?

we are currently displaying an error message on failure. This still needs to be dialed in further. Error messages range in value, some are more helpful than others(i.e. improper user/password)

- Edge case: how do we test + save for browser profiles that have 2fa via phone / email?

This is a tricky one because many 2fa applications require authentication for different purposes beyond just logging in. There does appear to be solutions available, would need to try them out for our specific use cases. See https://docs.browserbase.com/guides/authen

- Consideration: we should be careful with bot detection + repeated login attempts. We need to make sure our test has strict termination criteria so we don’t get a user locked out of their account

explicitly instructed agent not to retry on failure. Agent must stop immediately on 'invalid credentials' or 'account locked' errors. Only one login attempt per test to avoid triggering lockout policies

- UX visibility that might not be in scope: should we allow users to view / cancel / rerun the test for more control?

I like the idea of allowing the user to see the run, so they can confirm themself why the login might have failed. Cancel and rerun probably just come along with that functionality if we take it directly from the workflow portion

- UX flow that might not be in scope: when users add the same browser profile to multiple workflows, they might want to run things from workflow, to workflow, in a sequence because it is using the same account

When I was experimenting and testing, I found that the Learning Agent is actually smart enough to link proper credentials to a workflow, even if provided improper credentials. In action:

A Browser Profile has been created for www.login.com with user/password set

Another Browser Profile is created for www.login.com with fail/fail

The agent is smart enough to discard fail/fail and proceed to the proper user/password credential

I think in the same capacity, the agent should be able to string workflows together into a new workflow via the playwrite/other test <>code


skyvern/forge/sdk/routes/credentials.py
```diff
@@ -809,6 +739,382 @@ async def update_custom_credential_service_config(
         ) from e
 
 
+DEFAULT_LOGIN_PROMPT = (
```
celalzamanoglu: There's already a `DEFAULT_LOGIN_PROMPT` in `skyvern/forge/sdk/routes/run_blocks.py:42` that's reused by `workflow_copilot.py`. It handles cookie popups, 2FA steps, and termination on invalid credentials. Rather than defining a new one, you can import it

celalzamanoglu:
You can run `pre-commit run` in the `skyvern/` directory when you staged your changes, it'll run tests and pre-commit hooks on the changed (fe/be) files and fix the issues, so that your PR can pass the CI checks.

wintonzheng:
@jseres 
there are some pre-commit errors. could you run the precommit in your local and fix them?

skyvern/forge/sdk/workflow/models/block.py
```diff
@@ -3911,6 +3836,62 @@ class LoginBlock(BaseTaskBlock):
     # Parameter 1 of Literal[...] cannot be of type "Any"
     block_type: Literal[BlockType.LOGIN] = BlockType.LOGIN  # type: ignore
 
+    # Prefix injected into the navigation_goal when the browser already has
+    # session state (e.g. a reused browser session or a loaded browser profile).
+    # This tells the LLM to inspect the page for signs of an existing
+    # authenticated session BEFORE filling in any credentials.
+    _SESSION_CHECK_PREFIX: ClassVar[str] = (
```
wintonzheng: We already have a similar default prompt populated for user here:
https://github.com/Skyvern-AI/skyvern-cloud/blob/d59c6a10d3d416fc95d82e0ced1797b23253a56d/skyvern-frontend/src/routes/workflows/editor/nodes/LoginNode/types.ts#L28

we should just improve that prompt if that was not good enough

wintonzheng: @jseres curious what your htoughts are here about adding this prompt in the backend?

skyvern/forge/sdk/routes/credentials.py
```diff
@@ -809,6 +739,382 @@ async def update_custom_credential_service_config(
...
+    browser_profile_id = credential.browser_profile_id
+    browser_profile_url = credential.browser_profile_url
+    browser_profile_failure_reason: str | None = None
+
+    # If completed successfully and no browser profile yet, try to create one
+    if status == "completed" and not browser_profile_id:
+        workflow = await app.DATABASE.get_workflow(
+            workflow_id=workflow_run.workflow_id, organization_id=organization_id
+        )
+        if workflow and getattr(workflow, "persist_browser_session", False):
```
wintonzheng: this should be good right?
```suggestion
        if workflow and workflow.persist_browser_session:
```

skyvern/forge/sdk/routes/credentials.py
```diff
@@ -809,6 +739,382 @@ async def update_custom_credential_service_config(
...
+
+    # Create the workflow
+    workflow = await app.WORKFLOW_SERVICE.create_workflow_from_request(
+        organization=current_org,
+        request=workflow_create_request,
+    )
+
+    # Run the workflow — pass existing browser_profile_id so the browser launches
+    # with saved session data (cookies, storage) before the agent starts acting.
+    from skyvern.forge.sdk.workflow.models.workflow import WorkflowRequestBody  # noqa: PLC0415
```
wintonzheng: let's not do inline import

skyvern/forge/sdk/routes/credentials.py
```diff
@@ -809,6 +739,382 @@ async def update_custom_credential_service_config(
...
+        "created and linked to the credential."
+    ),
+    tags=["Credentials"],
+)
+@base_router.get(
+    "/credentials/{credential_id}/test/{workflow_run_id}/",
+    response_model=TestCredentialStatusResponse,
+    include_in_schema=False,
+)
+@legacy_base_router.get(
```
wintonzheng: no need to support legacy_base_router

skyvern/forge/sdk/routes/credentials.py
```diff
@@ -809,6 +739,382 @@ async def update_custom_credential_service_config(
...
+        "Optionally saves the browser profile after a successful login for reuse in workflows."
+    ),
+    tags=["Credentials"],
+)
+@base_router.post(
+    "/credentials/{credential_id}/test/",
+    response_model=TestCredentialResponse,
+    include_in_schema=False,
+)
+@legacy_base_router.post(
```
wintonzheng: no need to support legacy_base_router

skyvern/forge/sdk/routes/credentials.py
```diff
@@ -809,6 +739,382 @@ async def update_custom_credential_service_config(
...
+    workflow_run = await app.WORKFLOW_SERVICE.setup_workflow_run(
+        request_id=None,
+        workflow_request=run_request,
+        workflow_permanent_id=workflow.workflow_permanent_id,
+        organization=current_org,
+        max_steps_override=None,
+    )
+
+    # Execute the workflow in the background
+    background_tasks.add_task(
```
wintonzheng: let's use `await AsyncExecutorFactory.get_executor().execute_workflow`.

our cloud version would enqueue the job and run the workflow separately in a worker

skyvern/forge/sdk/routes/credentials.py
```diff
@@ -809,6 +739,382 @@ async def update_custom_credential_service_config(
...
+                            credential_id=credential_id,
+                            workflow_run_id=workflow_run_id,
+                            attempt=attempt + 1,
+                            max_retries=max_retries,
+                        )
+                        await asyncio.sleep(2)
+
+                if session_dir:
+                    # Create the browser profile in DB
+                    profile = await app.DATABASE.create_browser_profile(
```
wintonzheng: will it potentially create 5 browser profiles?

skyvern/forge/sdk/routes/credentials.py
suchintan: we should put this in CLAUDE.md

alembic/versions/2026_02_10_1200-add_browser_profile_to_credentials.py
```diff
@@ -0,0 +1,29 @@
...
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "a1b2c3d4e5f6"
```
AronPerez: This migration uses `revision = "a1b2c3d4e5f6"` and `down_revision = "43217e31df12"`, but there is already an existing migration at `alembic/versions/2026_02_13_0000-add_2fa_waiting_state_fields.py` with the **exact same revision ID** and the same `down_revision`. This will cause Alembic to fail with a duplicate revision error.

The new migration needs:
1. A unique revision ID (use `alembic revision --autogenerate` to generate one)
2. Its `down_revision` should chain off the actual latest migration head

skyvern/forge/sdk/routes/credentials.py
```diff
@@ -735,6 +739,382 @@ async def update_custom_credential_service_config(
...
+    if not workflow_run:
+        raise HTTPException(status_code=404, detail=f"Workflow run {workflow_run_id} not found")
+
+    status = str(workflow_run.status)
+    browser_profile_id = credential.browser_profile_id
+    browser_profile_url = credential.browser_profile_url
+    browser_profile_failure_reason: str | None = None
+
+    # If completed successfully and no browser profile yet, try to create one
+    if status == "completed" and not browser_profile_id:
```
AronPerez, line 1309: This `GET` status endpoint both reads and writes state (creating browser profiles, copying session data, updating credentials), which breaks REST expectations for a safe/idempotent `GET`. It also has a race where rapid polling can hit status == "completed" and not `browser_profile_id concurrently`, creating duplicate browser profiles. Creation should move to a one-shot async/background path, and the GET should be `read-only`

AronPerez (commented):
Thorough review of the Browser Profile Component PR. The feature design is solid — reusing existing workflow infrastructure for credential testing is a smart approach. However, there are several issues ranging from critical (migration collision, race conditions) to minor (naming, validation) that should be addressed before merge.

Several of these overlap with existing review comments (use `AsyncExecutorFactory`, drop `legacy_base_router`, move inline imports to top-level, use `workflow.persist_browser_session` directly). Those are not repeated in detail here but should also be addressed.

skyvern/forge/sdk/workflow/models/block.py
```diff
@@ -3819,6 +3836,62 @@ class LoginBlock(BaseTaskBlock):
...
+            if not self.navigation_goal.startswith("IMPORTANT — BEFORE"):
+                LOG.info(
+                    "LoginBlock: augmenting navigation_goal with session-check prefix",
+                    workflow_run_id=workflow_run_id,
+                    browser_session_id=browser_session_id,
+                    browser_profile_id=(
+                        workflow_run.browser_profile_id if workflow_run else None
+                    ),
+                )
+                self.navigation_goal = self._SESSION_CHECK_PREFIX + self.navigation_goal
```
AronPerez: This modifies the block instance's `navigation_goal` in-place. If the same `LoginBlock` instance is re-executed (retry, loop), the prefix is already there from the first run. The `startswith` guard prevents double-prepend within a single call, but modifying shared model state in an `execute()` method is a code smell.

Suggestion: Use a local variable for the augmented goal and pass it to the parent via a different mechanism, without mutating `self`:
```python
navigation_goal = self.navigation_goal
if has_existing_session and navigation_goal:
    navigation_goal = self._SESSION_CHECK_PREFIX + navigation_goal
# pass navigation_goal to super().execute() somehow
```

skyvern-frontend/src/routes/credentials/CredentialItem.tsx
```diff
@@ -5,6 +5,15 @@ import {
   isSecretCredential,
 } from "@/api/types";
 import { DeleteCredentialButton } from "./DeleteCredentialButton";
+import { CheckCircledIcon } from "@radix-ui/react-icons";
+
+function getHostname(url: string): string | null {
```
AronPerez: This is defined in both `CredentialItem.tsx` and `CredentialsModal.tsx` (with a minor return type difference: `string | null` vs `string`). Consider extracting to a shared utility, e.g., `src/util/getHostname.ts`.

skyvern-frontend/src/routes/credentials/CredentialsModal.tsx
```diff
@@ -101,6 +121,27 @@ function CredentialsModal({
     SECRET_CREDENTIAL_INITIAL_VALUES,
   );
 
+  // Test & Save Browser Profile state
+  const [testAndSave, setTestAndSave] = useState(false);
+  const [testUrl, setTestUrl] = useState("");
+  const [testStatus, setTestStatus] = useState<
+    "idle" | "testing" | "completed" | "failed" | "profile_failed"
+  >("idle");
+  const [, setTestWorkflowRunId] = useState<string | null>(null);
```
AronPerez: The state value is never read — only the setter is used. If the value is never consumed anywhere, this state is unnecessary; the `workflow_run_id` is already tracked locally in `startTest` and `pollTestStatus`.

skyvern-frontend/src/routes/credentials/CredentialsModal.tsx
```diff
@@ -126,26 +167,155 @@ function CredentialsModal({
...
+            description:
+              data.failure_reason ?? "The login test did not succeed",
+            variant: "destructive",
+          });
+          return;
+        }
+        // Still running — schedule next poll (no overlap possible)
+        pollIntervalRef.current = setTimeout(() => {
+          pollTestStatus(credentialId, workflowRunId);
+        }, 3000);
```
AronPerez: **Minor: Polling has no maximum retry count or backoff for network errors**

This `catch` block retries indefinitely with a flat 3-second interval. A network outage would cause infinite polling until the component unmounts. Consider adding a maximum retry count (e.g., 20 retries = ~60s) and/or exponential backoff.

skyvern-frontend/src/routes/credentials/CredentialsModal.tsx
```diff
@@ -187,6 +357,16 @@ function CredentialsModal({
         });
         return;
       }
+
+      if (testAndSave && testUrl.trim() === "") {
```
AronPerez: This only checks for a non-empty string. A user could enter `"hello"` and the backend would attempt to navigate to it. Consider adding a basic URL validation (e.g., check it starts with `https://` or `http://`).

tests/e2e/BROWSER_PROFILE_TEST_RESULTS.md
```diff
@@ -0,0 +1,166 @@
+# Browser Profile Test Results — DEFINITIVE PROOF
```
AronPerez: This file probably shouldn't be committed

skyvern/forge/sdk/schemas/credentials.py
```diff
@@ -161,6 +161,63 @@ class CredentialResponse(BaseModel):
     )
     credential_type: CredentialType = Field(..., description="Type of the credential")
     name: str = Field(..., description="Name of the credential", examples=["Amazon Login"])
+    browser_profile_id: str | None = Field(default=None, description="Browser profile ID linked to this credential")
+    browser_profile_url: str | None = Field(default=None, description="URL of the website the browser profile was created from")
```
AronPerez: This actually stores the login page URL used during the credential test, not a URL *to* the browser profile. Consider renaming to `login_url` or `tested_url` for clarity

jseres:
Merge conflicts resolved and looks like all comments have been addressed

wintonzheng:
Can we get rid of `.venv-test/` and `browser_sessions/`?

jseres:
> Can we get rid of `.venv-test/` and `browser_sessions/`?

oops, removed

PR #4747, opened 2026-02-13
Fix ForLoop script generation SyntaxError

pedrohsdb:
Synced from cloud PR: https://github.com/Skyvern-AI/skyvern-cloud/pull/8781
Author: @pedrohsdb

## Summary

- Fixes a bug where generated Python scripts containing ForLoop blocks fail with `SyntaxError: 'async for' outside async function`
- Root cause: `append_block_code()` placed bare `async for` statements at the module level of the generated script, but `async for` is only valid inside an `async def` function body
- The fix removes the `append_block_code()` call for ForLoop blocks — the for-loop code is already correctly inlined inside `run_workflow()` via `_build_block_statement()`
- The database `script_block` entry for cache tracking is unaffected (still created via `create_or_update_script_block()`)

## Why This Fix Is Necessary

Any workflow with a ForLoop block that generates a cached script produces invalid Python. When the system tries to load the script for `run_with: code` execution, `exec_module` hits the SyntaxError, and the workflow silently falls back to agent execution. This means **script caching is completely broken for ForLoop workflows** — they always run as agent, defeating the purpose of caching.

This was discovered during smoke test investigation of WPID `wpid_492390455756053710`:
1. Run 1 (agent) completed successfully and generated a cached script
2. Run 2 (code) found the cached script but failed to load it — `exec_module` hit `SyntaxError: 'async for' outside async function` at line 64 of main.py
3. The system fell back to agent execution

### How The Bug Manifests

The generated script had this structure:
```python
# Inside run_workflow() — CORRECT:
async def run_workflow(parameters):
 async for current_value in skyvern.loop(...):
     await skyvern.extract(...)

# At module level — INVALID (causes SyntaxError):
async for current_value in skyvern.loop(...):
 await skyvern.extract(...)
```

The duplicate came from `append_block_code(block_code)` which adds code to the module-level `block_fns` list. Task blocks add **function definitions** there (valid at module level), but ForLoop blocks added bare **`async for` statements** (invalid at module level).

### Unit Tests
- [x] Added regression test `test_forloop_script_compiles_without_syntax_error` that generates a full script with a ForLoop block via `generate_workflow_script_python_code()` and verifies it compiles via `ast.parse()`
- [x] All 13 ForLoop tests pass (`tests/unit/test_forloop_script_generation.py`)
- [x] All 17 conditional caching tests pass (`tests/unit/test_conditional_script_caching.py`)

### End-to-End Local Test
Ran a full end-to-end test using `dev_scripts/run_workflow_from_json.py` with a ForLoop workflow that iterates over 2 URLs and extracts page titles:

| | Run 1 | Run 2 |
|---|---|---|
| **run_with** | `agent` | `code` |
| **Status** | completed | completed |
| **Script generated?** | Yes | No (reused cached) |

- **Run 1**: Agent mode executed the ForLoop, generated a cached script. Verified the `main.py` artifact only has `async for` inside `run_workflow()` (no module-level duplicate)
- **Run 2**: `--run-with code` found the cached script, loaded it successfully (no SyntaxError), executed with `run_with=code`, completed. Logs confirmed: `"Found cached script for workflow"` and `"Workflow script already up to date; skipping regeneration"`
- Database confirmed: both runs completed, Run 1 `run_with=agent`, Run 2 `run_with=code`, no new scripts generated on Run 2

### DIFF
--- a/tests/unit/test_forloop_script_generation.py
+++ b/tests/unit/test_forloop_script_generation.py
@@ -6,6 +6,7 @@ These tests verify that ForLoop blocks are properly handled during:
 2. Script generation (generate_script.py)
 """
 
+import ast
 from typing import Any
 from unittest.mock import AsyncMock, MagicMock, patch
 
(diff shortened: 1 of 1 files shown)


suchintan:
Merged here: https://github.com/Skyvern-AI/skyvern/pull/4755

PR #4750, opened 2026-02-14
Fix misleading error message when LLM returns empty actions

pedrohsdb:
Synced from cloud PR: https://github.com/Skyvern-AI/skyvern-cloud/pull/8784
Author: @pedrohsdb

## Summary

- Fixes a misleading error message that blamed "LLM service errors" when the LLM was actually working fine but found nothing to interact with on the page

## Context

When a step fails because the LLM returned an empty actions list (e.g., a conditional popup didn't appear, or there were no matching elements), the agent retries until `max_retries` is exhausted. The `steps_without_actions` counter was then used to generate the failure reason — but it conflated two distinct cases:

1. **Actual LLM failure** (`cur_step.output is None`): The LLM call itself errored — rate limiting, outage, etc.
2. **Empty actions** (`cur_step.output` is set, but `actions_and_results` is empty): The LLM worked correctly, analyzed the page, and determined there was nothing to click/fill/interact with.

Both were counted the same way (`not cur_step.output or not cur_step.output.actions_and_results`), and both got the message: *"This is typically caused by LLM service errors during action extraction, such as rate limiting, service outages, or resource exhaustion."*

This confused customers (e.g., Signal Advisors) into thinking the AI was broken, when the real issue was that the expected page content wasn't present.

## Fix

Narrowed the `steps_without_actions` condition to only count true LLM failures (`not cur_step.output`). Steps where the LLM returned empty actions now fall through to the normal summarization path, which calls the secondary LLM to generate a contextual explanation of why the task actually failed — a much more accurate and helpful message.

## Test plan

- [x] `py_compile` passes
- [x] All 497 unit tests pass
- [ ] Verify in staging: trigger a task where a conditional element isn't found → confirm the failure message no longer mentions "LLM service errors"

### DIFF
--- a/tests/unit/test_forloop_script_generation.py
+++ b/tests/unit/test_forloop_script_generation.py
@@ -6,6 +6,7 @@ These tests verify that ForLoop blocks are properly handled during:
 2. Script generation (generate_script.py)
 """
 
+import ast
 from typing import Any
 from unittest.mock import AsyncMock, MagicMock, patch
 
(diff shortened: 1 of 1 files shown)


suchintan:
Merged here: https://github.com/Skyvern-AI/skyvern/pull/4755

PR #4809, opened 2026-02-19
Add API endpoint to clear cached scripts for workflows

suchintan:
This adds a DELETE /scripts/{workflow_permanent_id}/cache endpoint that allows clearing all cached scripts for a specific workflow programmatically. The endpoint soft-deletes database entries and clears the in-memory TTL cache.

This addresses the need for users to bulk-clear cached scripts without manually deleting them one-by-one, which is especially useful when cached code contains stale PII that needs to be purged to trigger regeneration.

Slack thread: https://skyvern.slack.com/archives/C0922S2MVJ8/p1771526657580269?thread_ts=1771008685.387109&cid=C0922S2MVJ8

https://claude.ai/code/session_01JcWzFaBkZ6YdzVacfuTpGm

### DIFF
--- a/skyvern/forge/sdk/routes/scripts.py
+++ b/skyvern/forge/sdk/routes/scripts.py
@@ -622,7 +622,7 @@ async def clear_workflow_cache(
     )
 
     # Clear in-memory cache
-    workflow_script_service.clear_workflow_script_cache(
+    cache_cleared_count = workflow_script_service.clear_workflow_script_cache(
         organization_id=current_org.organization_id,
         workflow_permanent_id=workflow_permanent_id,
     )
@@ -632,9 +632,10 @@ async def clear_workflow_cache(
         organization_id=current_org.organization_id,
         workflow_permanent_id=workflow_permanent_id,
         deleted_count=deleted_count,
+        cache_cleared_count=cache_cleared_count,
     )
 
     return ClearCacheResponse(
         deleted_count=deleted_count,
-        message=f"Successfully cleared {deleted_count} cached script(s) for workflow {workflow_permanent_id}",
+        message=f"Successfully cleared {deleted_count} database record(s) and {cache_cleared_count} in-memory cache entry(s) for workflow {workflow_permanent_id}",
     )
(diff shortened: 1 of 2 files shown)


skyvern/services/workflow_script_service.py
```diff
@@ -57,6 +57,41 @@ def _make_workflow_script_cache_key(
...
+    """
+    keys_to_delete = []
+
+    for key in _workflow_script_cache.keys():
+        # Key format: (org_id, workflow_permanent_id, cache_key_value, workflow_run_id, cache_key, statuses_key)
+        if len(key) >= 2 and key[0] == organization_id:
+            if workflow_permanent_id is None or key[1] == workflow_permanent_id:
+                keys_to_delete.append(key)
+
+    for key in keys_to_delete:
```
AronPerez, line 82: Iterating over `_workflow_script_cache.keys()` while other async handlers can mutate the cache (e.g., `get_workflow_script_by_cache_key_value` writes to it) risks a `RuntimeError: dictionary changed size during iteration`. Snapshot the keys first:
```suggestion
    for key in list(_workflow_script_cache.keys()):
```
This doesn't fully eliminate the race between the snapshot and the subsequent `.pop()` calls, but it prevents the iteration crash. If the cache sees heavy concurrent access, a lock around the whole clear operation would be safer.

skyvern/forge/sdk/routes/scripts.py
```diff
@@ -573,3 +574,67 @@ async def delete_workflow_cache_key_value(
...
+
+    LOG.info(
+        "Cleared workflow cache",
+        organization_id=current_org.organization_id,
+        workflow_permanent_id=workflow_permanent_id,
+        deleted_count=deleted_count,
+    )
+
+    return ClearCacheResponse(
+        deleted_count=deleted_count,
```
AronPerez, line 639: `deleted_count` only reflects the database soft-delete count. The in-memory cache may hold a different number of entries, so the message `"Successfully cleared {deleted_count} cached script(s)"` can be misleading. Consider including the in-memory count returned by `clear_workflow_script_cache()` or clarifying the message to say "database" entries.

wintonzheng:
@claude can you check @AronPerez 's comments? are they legit and good to address? if yes make the change

wintonzheng (approved):

PR #4811, opened 2026-02-19
[SKY-7980] Patch Credential TOTP Over Webhook Logic

AronPerez:
## Ticket  

[SKY-7980]

## Description 

Commit 8714f15f ("Fix MFA resolution priority: credential TOTP over webhook") restructured the verification code handling to try credential TOTP first, then fall back to poll_otp_value(). However, it accidentally gated the

poll_otp_value() call behind (task.totp_verification_url or task.totp_identifier), which means polling never starts when no TOTP credentials are configured.

This breaks the manual 2FA code submission flow — poll_otp_value() is what sets waiting_for_verification_code=True in the DB, publishes the WebSocket notification, and polls for manually submitted codes. Without it being

called, the frontend never shows the 2FA input prompt.

poll_otp_value() already handles the no-config case correctly via its else branch (line 222-228 of otp_service.py) which calls _get_otp_value_by_run().

## How Has This Been Tested?  

1. Set VITE_ENABLE_2FA_NOTIFICATIONS=true in skyvern-frontend/.env
2. skyvern run all
3. Create workflow that visits https://authenticationtest.com/totpChallenge/, has OTP challenge
4. Run workflow
5. Get TOTP poll notif

```shell
2026-02-19T19:53:51.283805Z [info     ] [agent.py                      :4528] Need verification code         [skyvern.forge.agent]
/home/devuser/skyvern/skyvern/services/otp_service.py:112: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
  start_datetime = datetime.utcnow()
2026-02-19T19:53:51.297874Z [info     ] [otp_service.py                :118 ] Polling otp value              [skyvern.services.otp_service] task_id=tsk_497561556493169102 workflow_run_id=wr_497561552198201800 workflow_permanent_id=wpid_496471078926911466 totp_verification_url=None totp_identifier=                                                                                                                                                              
2026-02-19T19:53:51.305723Z [info     ] [otp_service.py                :137 ] Set 2FA waiting state for workflow run [skyvern.services.otp_service] workflow_run_id=wr_497561552198201800 verification_code_identifier=
/home/devuser/skyvern/skyvern/forge/sdk/routes/streaming/screenshot.py:182: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
  
``` 

### Validating Workflows With totp_verification_url Get TOTP

https://github.com/user-attachments/assets/f071d152-5895-4636-9d76-6a95a971f974

### Validating Workflows Without totp_verification_url Get TOTP

https://github.com/user-attachments/assets/4e05041c-f6c2-4d49-b95c-c50eb2d54dae


wintonzheng (approved):

skyvern/forge/agent.py
```diff
@@ -4529,7 +4529,7 @@ async def handle_potential_verification_code(
             # Try credential TOTP first (highest priority, doesn't need totp_url/totp_identifier)
             otp_value = try_generate_totp_from_credential(task.workflow_run_id)
             # Fall back to webhook/totp_identifier
-            if not otp_value and (task.totp_verification_url or task.totp_identifier):
+            if not otp_value:
```
AronPerez, line 4532: The old guard `(task.totp_verification_url or task.totp_identifier)` was the root cause — it prevented `poll_otp_value()` from ever being called in the manual 2FA flow.

This fix is correct: `poll_otp_value()` already handles the no-config case via its `else` branch (otp_service.py:222-228) which calls `_get_otp_value_by_run()`. Removing the guard lets that path execute.

Nit: the comment on line 4531 still says `"Fall back to webhook/totp_identifier"` — now that this also covers the manual submission path, consider updating it to something like `"Fall back to webhook / totp_identifier / manual submission"`.

skyvern/webeye/actions/parse_actions.py
```diff
@@ -916,7 +916,7 @@ async def generate_cua_fallback_actions(
         # Try credential TOTP first (highest priority, doesn't need totp_url/totp_identifier)
         otp_value = try_generate_totp_from_credential(task.workflow_run_id)
         # Fall back to webhook/totp_identifier
-        if not otp_value and (task.totp_verification_url or task.totp_identifier) and task.organization_id:
+        if not otp_value and task.organization_id:
```
AronPerez, line 919: Same fix as in `agent.py` — good that both call sites were updated consistently.

Note that this path still requires `task.organization_id` (which is correct — `poll_otp_value` needs an org token). The agent.py path has the same guard at line 4527. Consistent behavior.

skyvern/forge/sdk/routes/streaming/notifications.py
```diff
@@ -22,6 +22,23 @@ async def notification_stream(
     websocket: WebSocket,
     apikey: str | None = None,
     token: str | None = None,
+) -> None:
+    return await _notification_stream_handler(websocket=websocket, apikey=apikey, token=token)
+
+
+@legacy_base_router.websocket("/stream/notifications")
```
AronPerez, line 29: Adding the legacy router endpoint here is necessary for the 2FA WebSocket notification to work when the frontend connects via the legacy base path.

Good refactor: extracting `_notification_stream_handler` avoids duplicating the full handler body across both routers.

skyvern/forge/sdk/routes/streaming/notifications.py
```diff
@@ -22,6 +22,23 @@ async def notification_stream(
     websocket: WebSocket,
     apikey: str | None = None,
     token: str | None = None,
+) -> None:
+    return await _notification_stream_handler(websocket=websocket, apikey=apikey, token=token)
```
AronPerez, line 26: Minor: both `notification_stream` and `notification_stream_legacy` use `return await ...`. Since the return type is `None`, the `return` is unnecessary (though harmless). Could just be `await _notification_stream_handler(...)` for consistency with typical FastAPI endpoint style.

tests/unit/test_credential_totp_priority.py
```diff
@@ -165,3 +166,135 @@ def test_skips_non_dict_values(self) -> None:
...
+        mock_task.organization_id = "org_123"
+        mock_task.task_id = "task_456"
+        mock_task.workflow_run_id = None
+        mock_task.navigation_goal = "test goal"
+
+        mock_step = MagicMock()
+        mock_step.step_id = "step_789"
+        mock_step.order = 0
+
+        mock_otp_value = MagicMock(spec=OTPValue)
```
AronPerez, line 192: This test is the key regression guard: it validates that `poll_otp_value` is called even when both `totp_verification_url` and `totp_identifier` are `None` — the exact scenario the bug broke.

Consider adding a brief docstring-level note that this tests the *manual 2FA code submission* flow (where the frontend prompts the user), since that is the user-facing behavior being protected.

tests/unit/test_credential_totp_priority.py
```diff
@@ -165,3 +166,135 @@ def test_skips_non_dict_values(self) -> None:
...
+            from skyvern.webeye.actions.actions import VerificationCodeAction
+
+            assert isinstance(actions[0], VerificationCodeAction)
+            assert actions[0].verification_code == "123456"
+
+    @pytest.mark.asyncio
+    async def test_poll_otp_called_when_no_totp_config_agent(self) -> None:
+        """When credential TOTP returns None and no totp_verification_url or
+        totp_identifier is set, poll_otp_value should still be called via the
+        agent's handle_potential_verification_code so the manual code submission
```
AronPerez, line 245: This instantiation pattern (`ForgeAgent.__new__(ForgeAgent)`) bypasses `__init__`, which works here because the test only calls `handle_potential_verification_code` and all its dependencies are mocked.

This is fine for a targeted unit test but is fragile — if `handle_potential_verification_code` ever accesses `self.some_field`, this will break with an `AttributeError`. A comment noting why `__new__` is used (to avoid constructor dependencies) would help future readers.

tests/unit/test_credential_totp_priority.py
```diff
@@ -165,3 +166,135 @@ def test_skips_non_dict_values(self) -> None:
...
+                "place_to_enter_verification_code": True,
+                "should_enter_verification_code": True,
+            }
+
+            result = await agent.handle_potential_verification_code(
+                task=mock_task,
+                step=MagicMock(),
+                scraped_page=MagicMock(),
+                browser_state=MagicMock(),
+                json_response=json_response,
```
AronPerez, line 284: Good: the assertion verifies the exact kwargs passed to `poll_otp_value`, including `workflow_id=None` and `workflow_permanent_id=None`. This matches the code path where `task.workflow_run_id` is `None` so the workflow lookup is skipped.

AronPerez:
@claude merge this

PR #4818, opened 2026-02-19
feat: Browser profile testing & login-free workflow support — backend (#SKY-8012)

celalzamanoglu:
Synced from cloud PR: https://github.com/Skyvern-AI/skyvern-cloud/pull/8819
Author: @celalzamanoglu

## Summary - [SKY-8012](https://linear.app/skyvern/issue/SKY-8012)

Backend implementation for credential browser profile testing and login-free workflow support. This PR adds the ability to test a saved credential by running a real login workflow, capture the resulting authenticated browser session as a reusable browser profile, and then automatically load that profile in future workflow runs that contain a login block — so the session is restored instead of logging in from scratch every time.

## Problem

Workflows that interact with authenticated sites have to log in at the start of every run. This is slow, fragile, and risks account lockouts from repeated login attempts. The browser profile feature solves this by letting users verify their credentials once, capture the resulting authenticated session, and reuse it across runs — bypassing the login step entirely when the session is still valid, and falling back gracefully to a real login when it has expired.

## What Changed

**New API endpoints (`skyvern/forge/sdk/routes/credentials.py`):**
- `POST /credentials/test-login` — test login with inline credentials (no saved credential required); creates a temporary credential, spins up a login-only workflow, and returns a `workflow_run_id` to poll
- `POST /credentials/{id}/test` — test an existing saved credential against a given URL
- `GET /credentials/{id}/test/{workflow_run_id}` — poll the status of a running test; returns humanized failure messages and exposes `browser_profile_id` and `browser_profile_failure_reason` once the background task completes
- `POST /credentials/{id}/test/{workflow_run_id}/cancel` — cancel a running credential test and clean up temporary resources; returns `cancel_failed` instead of masking errors when cancellation cannot be confirmed
- `PATCH /credentials/{id}` — rename a credential and optionally record `tested_url`; fixes a pre-existing bug where the DB layer referenced a non-existent `website_url` column instead of `tested_url`

**Background browser profile creation (`_create_browser_profile_after_workflow`):**
- Polls the workflow run until it reaches a terminal state (up to 10 minutes at 5-second intervals)
- Retrieves the persisted browser session from storage with up to 5 retries to handle async persistence lag
- Creates a `BrowserProfile` DB record, copies session data to the profile storage location, and links `browser_profile_id` + `tested_url` back to the credential
- Cleans up temporary credentials (created by `test-login`) on failure, timeout, or cancellation
- Uses a module-level `_background_tasks` set to hold strong GC-safe references to asyncio tasks

**Login block browser profile integration (`skyvern/forge/sdk/workflow/service.py`):**
- Before executing a login block, resolves whether the block's credential has an associated browser profile by inspecting both `CredentialParameter` and `WorkflowParameter` (CREDENTIAL_ID type) styles
- When a browser profile is found, sets it on the workflow run, opens a browser with the saved session, and navigates to the block's URL
- Waits for `networkidle` (10-second timeout, non-fatal) before the agent acts so cookie-based redirects can complete
- Modifies the block's `navigation_goal` at runtime to instruct the AI to check if it is already logged in rather than unconditionally logging in; guards against navigation goal corruption on retries by saving the original goal before mutation
- Falls back to the normal login path (clears the profile from the workflow run, runs the block unmodified) if the browser profile fails to load
- Validates that the resolved browser profile still exists in the DB before using it; warns and skips if not found

**LLM navigation goal prompts:**
- `DEFAULT_LOGIN_PROMPT` — strict one-attempt-only login instructions with explicit termination criteria for wrong-password and unsupported-field scenarios
- `BROWSER_PROFILE_LOGIN_PROMPT` — variant used when a profile is loaded; instructs the AI to first check for an existing session before attempting login
- `LOGIN_TEST_TERMINATE_CRITERION` — termination condition passed to the login-test workflow to prevent retrying failed logins

**Humanized error messages (`_humanize_test_failure`):**
- Maps raw internal failure reasons (wrong password, 2FA, navigation errors, timeouts) to user-friendly messages for the polling endpoint response

**Schema changes (`skyvern/forge/sdk/schemas/credentials.py`):**
- Added `browser_profile_id` and `tested_url` fields to `CredentialResponse` and `Credential`
- Added `UpdateCredentialRequest`, `TestCredentialRequest`, `TestLoginRequest`, `TestCredentialResponse`, `TestLoginResponse`, `TestCredentialStatusResponse`, `CancelTestResponse`
- SSRF protection via `validate_url` model validators on `TestCredentialRequest` and `TestLoginRequest`

**Database layer (`skyvern/forge/sdk/db/agent_db.py`, `models.py`):**
- `CredentialModel` gets `browser_profile_id` and `tested_url` columns
- `update_credential` fixed to use `_UNSET` sentinel pattern for optional fields (`browser_profile_id`, `tested_url`) and filters out soft-deleted rows
- `update_workflow_run` gains a `browser_profile_id` parameter (also `_UNSET` sentinel) so the login block integration can set or clear the profile on a running workflow run

**Alembic migration (`alembic/versions/2026_02_19_1947-0327421609f9_add_browser_profile_to_credentials.py`):**
- Adds `browser_profile_id` (nullable string) and `tested_url` (nullable string) columns to the `credentials` table

**Block model (`skyvern/forge/sdk/workflow/models/block.py`):**
- `BaseTaskBlock` waits for `networkidle` after opening a browser with a loaded profile, before the agent begins interacting with the page

**Workflow schema (`skyvern/schemas/workflows.py`):**
- Added `skipped` to `BlockStatus` enum for future use when login blocks are fully bypassed

**Cloud init (`cloud/__init__.py`):**
- Wires the pooled Redis client into `NotificationRegistryFactory` so the credential test polling infrastructure uses the shared connection pool

## Test plan
- [ ] Run the Alembic migration on a local or staging database and verify both columns are added to `credentials`
- [ ] Call `POST /api/v1/credentials/test-login` with a valid username, password, and URL; verify a `credential_id` and `workflow_run_id` are returned and the workflow starts running
- [ ] Poll `GET /api/v1/credentials/{id}/test/{wr_id}` and verify the status transitions from `running` to `completed`; after ~30 seconds past completion, verify `browser_profile_id` is populated on the response
- [ ] Call `POST /api/v1/credentials/test-login` with invalid credentials; verify the poll response contains a humanized `failure_reason` and no `browser_profile_id`
- [ ] Call `POST /api/v1/credentials/{id}/test/{wr_id}/cancel` while a test is running; verify it returns `{"status": "canceled"}` and the temporary credential is deleted
- [ ] Run a workflow with a login block whose credential has a `browser_profile_id`; verify the block loads the profile, navigates to the URL, and completes without performing a new login
- [ ] Run the same workflow after the saved session has expired; verify the login block falls back to performing a real login
- [ ] Delete the browser profile from the database while a credential still references it; verify the login block logs a warning and falls back to normal login rather than erroring
- [ ] Call `PATCH /api/v1/credentials/{id}` with a new name and `tested_url`; verify both fields are updated
- [ ] Test `POST /api/v1/credentials/test-login` with a private or localhost URL (SSRF attempt); verify a 400 is returned


celalzamanoglu:
⚠️ **Migration Alert**

The source cloud PR included database migration changes in `alembic/versions/` that are **NOT automatically synced**.

If these migrations affect open-source tables, please manually create corresponding migrations in the OSS repo.

Source PR: https://github.com/Skyvern-AI/skyvern-cloud/pull/8819

PR #4825, opened 2026-02-20
show version number in frontend

wintonzheng:


Dockerfile.ui
```diff
@@ -1,9 +1,27 @@
 FROM node:20.12-slim
 
-# Install tini for proper signal handling and zombie reaping
-RUN apt-get update && apt-get install -y --no-install-recommends tini && rm -rf /var/lib/apt/lists/*
+# Install tini and git for proper signal handling and version info
+RUN apt-get update && apt-get install -y --no-install-recommends tini git && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /app
+
+# Copy necessary files for version extraction
```
AronPerez, line 8: 🔴 **Security: `.git` directory leaked into Docker image layers**

This `COPY .git /tmp/.git` adds the entire git history into an image layer. Even though `rm -rf /tmp/.git` runs later, Docker layers are immutable — the `.git` directory is fully recoverable from the image history by anyone with access to the image. This can expose commit history, author emails, and any secrets ever committed.

Note: `.dockerignore` already excludes `.git`, so this line explicitly bypasses that protection.

**Fix:** Extract git info on the host and pass it as build arguments instead:
```dockerfile
ARG GIT_SHA=unknown
ARG BUILD_TIME=unknown
ARG APP_VERSION=unknown
```
Then build with:
```bash
docker build \
  --build-arg GIT_SHA=$(git rev-parse --short=7 HEAD) \
  --build-arg BUILD_TIME=$(date -u +"%Y-%m-%dT%H:%M:%SZ") \
  --build-arg APP_VERSION=$(grep SDK_VERSION skyvern-ts/client/src/version.ts | sed -n 's/.*"\([^"]*\)".*/\1/p') \
  -f Dockerfile.ui .
```

Dockerfile.ui
```diff
@@ -1,9 +1,27 @@
...
+COPY ./skyvern-ts/client/src/version.ts /tmp/sdk-version.ts
+COPY .git /tmp/.git
+
```
AronPerez, line 11: 🟡 **Bug: Greedy sed regex may extract the wrong version string**

The `.*` capture group is greedy — it matches from the first `"` to the **last** `"` on the line. If `version.ts` ever has a trailing comment containing quotes (e.g. `// previously "1.0.18"`), the wrong string will be extracted.

**Fix:** Use a non-greedy character class:
```bash
sed -n 's/.*"\([^"]*\)".*/\1/p'
```

PR #4846, opened 2026-02-22
feat: fix issue 4019 custom proxy

### LINKED ISSUE #4019: Send proxy parameters along with the session
Hi everyone! We're self-hosting Skyvern but facing issues with proxies. We'd be very happy to pay for your time if you've faced this before and could help us solve it!

We are currently experiencing difficulties configuring an environment where we can send proxy parameters along with the session. Specifically, we need to manually define the proxy on a per-session basis through the app request, but we have not yet been able to get this setup to work as expected.

navanpreet:
- Updated ProxyLocationInput to accept raw proxy URI strings (e.g. http://user:pass@host:port)
- Modified serialization and parsing logic to accept custom proxy strings that fail Enum validation
- Updated Playwright browser factory to inject custom proxy credentials directly into the browser context
- Fixes #4019


skyvern/webeye/browser_factory.py
```diff
@@ -254,12 +254,25 @@ def build_browser_args(
...
+
+        if custom_proxy_config:
+            args["proxy"] = custom_proxy_config
+        elif settings.ENABLE_PROXY:
             proxy_config = setup_proxy()
             if proxy_config:
                 args["proxy"] = proxy_config
 
-        if proxy_location:
+        if proxy_location and not custom_proxy_config:
```
AronPerez: **Bug: raw string passed to  which only accepts  enum**

When  is a string that failed ,  is , so the condition  is  and the raw string falls through here.  has signature  and uses enum equality checks — passing a plain string means all comparisons silently return `False` (no timezone is set, no error raised), which may be acceptable but is unintentional.

If a valid enum value *happens* to match the string (unlikely), it would also silently pass type-checking.

**Suggested fix:**
```python
if proxy_location and not custom_proxy_config and isinstance(proxy_location, ProxyLocation):
    if tz_info := get_tzinfo_from_proxy(proxy_location=proxy_location):
        args["timezone_id"] = tz_info.key
```

skyvern/forge/sdk/db/utils.py
```diff
@@ -129,9 +129,11 @@ def _deserialize_proxy_location(value: str | None) -> ProxyLocationInput:
         result = ProxyLocation(value)
         return result
     except ValueError:
-        # If all else fails, return as-is (shouldn't happen with valid data)
-        LOG.warning("Failed to deserialize proxy_location", db_value=value)
-        return None
+        # If all else fails, return as-is (custom proxy URL)
+        LOG.warning(
+            "Failed to deserialize proxy_location as enum, assuming it is a custom proxy url string", db_value=value
```
AronPerez: **Security: proxy credentials may be leaked to logs**

When  is a custom proxy URL like , the full string including the username and password is written to the log.

**Suggested fix** — sanitize the URL before logging:
```python
from urllib.parse import urlparse

def _sanitize_proxy_url(url: str) -> str:
    try:
        p = urlparse(url)
        if p.username or p.password:
            netloc = f"<redacted>@{p.hostname}" + (f":{p.port}" if p.port else "")
            return url.replace(p.netloc, netloc)
    except Exception:
        pass
    return url

LOG.warning(
    "Failed to deserialize proxy_location as enum, assuming it is a custom proxy url string",
    db_value=_sanitize_proxy_url(value),
)
```

skyvern/webeye/browser_factory.py
```diff
@@ -254,12 +254,25 @@ def build_browser_args(
...
+                "server": f"{parsed.scheme}://{parsed.hostname}" + (f":{parsed.port}" if parsed.port else ""),
+            }
+            if parsed.username:
+                custom_proxy_config["username"] = parsed.username
+            if parsed.password:
+                custom_proxy_config["password"] = parsed.password
+
+        if custom_proxy_config:
+            args["proxy"] = custom_proxy_config
+        elif settings.ENABLE_PROXY:
```
AronPerez, line 330: **UX: invalid custom proxy string silently replaced by the settings proxy**

When  is a string but  returns ,  stays  and execution falls through to this branch. If  is , the user ends up using a completely different proxy without any warning — they believe their custom proxy is active when it is not.

**Suggested fix** — emit a warning when a string proxy fails validation:
```python
if proxy_location and isinstance(proxy_location, str):
    if _is_valid_proxy_url(proxy_location):
        # ... build custom_proxy_config ...
    else:
        LOG.warning("Invalid custom proxy URL, ignoring", proxy_location=proxy_location)
```

navanpreet:
Thanks for the review @AronPerez - I've addressed the feedback in the latest push:

- Bug Fix: Added isinstance(proxy_location, ProxyLocation) checks in browser_factory.py to ensure get_tzinfo_from_proxy is only called with enum instances, preventing silent failures when custom proxy strings
     are used.
- Security: Implemented _sanitize_proxy_url in db/utils.py to redact credentials from custom proxy URLs before they are logged.
- UX: Added a warning log for invalid custom proxy URLs in browser_factory.py to avoid silent fallback to the settings-defined proxy.

I've also verified the changes locally using ruff and mypy.

AronPerez (commented):
Review from code review session. Overall: custom proxy support is a good feature, but a few items need attention before merge.

skyvern/forge/sdk/db/utils.py
```diff
@@ -129,9 +129,24 @@ def _deserialize_proxy_location(value: str | None) -> ProxyLocationInput:
         result = ProxyLocation(value)
         return result
     except ValueError:
-        # If all else fails, return as-is (shouldn't happen with valid data)
-        LOG.warning("Failed to deserialize proxy_location", db_value=value)
-        return None
+        # If all else fails, return as-is (custom proxy URL)
+        from urllib.parse import urlparse
```
AronPerez: 🔴 **Critical**: Per project style guide: "Always place `import` and `from ... import` statements at the top of the file, not inline inside functions." Move `from urllib.parse import urlparse` to the module-level imports.

skyvern/forge/sdk/db/utils.py
```diff
@@ -129,9 +129,24 @@ def _deserialize_proxy_location(value: str | None) -> ProxyLocationInput:
...
+
```
AronPerez: 🟡 **Suggestion**: `_sanitize_proxy_url` is defined as a nested function inside the except block, meaning it gets re-created on every call. Move it to module level — it has no closure dependencies.

skyvern/schemas/runs.py
```diff
@@ -236,7 +236,7 @@ def validate_subdivision(cls, v: str | None) -> str | None:
 
 
 # Type alias for proxy location that accepts either legacy enum or new GeoTarget
```
AronPerez, line 251: 🟡 **Suggestion**: Adding bare `str` to the union means *any* string is accepted (e.g., `"hello"` would pass Pydantic validation). Consider adding a Pydantic validator that checks strings look like proxy URLs (scheme + host at minimum) to fail fast with a clear error, rather than silently failing at browser launch time with an "Invalid custom proxy URL, ignoring" warning.

skyvern/webeye/browser_factory.py
```diff
@@ -254,12 +254,28 @@ def build_browser_args(
         if settings.BROWSER_LOCALE:
             args["locale"] = settings.BROWSER_LOCALE
 
-        if settings.ENABLE_PROXY:
+        custom_proxy_config = None
```
AronPerez, line 314: 🟡 **Suggestion**: This custom proxy parsing logic duplicates what `_get_proxy_server_creds()` (line 417) already does. The difference is this version correctly separates `server` from credentials for Playwright's proxy API, which is actually more correct. Consider either:
1. Reusing `_get_proxy_server_creds` and constructing the server URL separately, or
2. Adding a comment explaining why the existing helper wasn't reused (the `server` field semantics differ).

skyvern/webeye/browser_factory.py
```diff
@@ -254,12 +254,28 @@ def build_browser_args(
...
+                }
+                if parsed.username:
+                    custom_proxy_config["username"] = parsed.username
+                if parsed.password:
+                    custom_proxy_config["password"] = parsed.password
+            else:
+                LOG.warning("Invalid custom proxy URL, ignoring", proxy_location=proxy_location)
+
+        if custom_proxy_config:
+            args["proxy"] = custom_proxy_config
```
AronPerez, line 329: 💡 **Question**: When a user passes a custom proxy string, the `ENABLE_PROXY` / `setup_proxy()` path is skipped entirely (custom proxy takes precedence via the `elif`). Is this the intended behavior? The cloud layer may have billing or routing logic tied to the hosted proxy pool. Worth verifying that custom proxy users won't silently bypass proxy billing in the cloud deployment.

skyvern/schemas/browser_sessions.py
```diff
@@ -18,7 +18,7 @@ class CreateBrowserSessionRequest(BaseModel):
         ge=MIN_TIMEOUT,
         le=MAX_TIMEOUT,
     )
-    proxy_location: ProxyLocation | None = Field(
+    proxy_location: ProxyLocationInput | None = Field(
```
AronPerez: 🟢 **Nice**: Good change — `CreateBrowserSessionRequest` now accepts `ProxyLocationInput` instead of just `ProxyLocation`, making the API consistent with task/workflow run requests.

skyvern/forge/sdk/db/utils.py
navanpreet: Done. Moved to module-level imports and promoted _sanitize_proxy_url to module-level function

skyvern/forge/sdk/db/utils.py
navanpreet: Addressed

skyvern/schemas/runs.py
navanpreet: Good call. Added a validate_proxy_location_str() helper that checks for a valid proxy scheme and a hostname

skyvern/webeye/browser_factory.py
navanpreet: Added a comment explaining the distinction - Playwright's proxy API requires server to be scheme://host:port only (no credentials embedded), while _get_proxy_server_creds returns the full URL as server

skyvern/webeye/browser_factory.py
navanpreet: Yes, this is intentional — custom proxy users are bringing their own proxy, so they should bypass the hosted pool entirely. Added a comment in the code making this explicit. Billing implications for cloud deployment are a good callout but would be handled at the API/authorization layer rather than in the browser factory

skyvern/schemas/runs.py
```diff
@@ -241,7 +241,25 @@ def validate_subdivision(cls, v: str | None) -> str | None:
...
+        from urllib.parse import urlparse
+
+        parsed = urlparse(value)
+        if parsed.scheme in _PROXY_URL_SCHEMES and parsed.hostname:
+            return value
+    except Exception:
+        pass
+    raise ValueError(
+        f"Invalid proxy URL: must start with http://, https://, or socks5:// and include a hostname. Got: {value!r}"
+    )
```
AronPerez: Two separate proxy URL validation functions now exist `validate_proxy_location_str() in schemas/runs.py` and `_is_valid_proxy_url() in webeye/browser_factory.py`. Let's stick with `_is_valid_proxy_url` and make it available across modules

skyvern/forge/sdk/schemas/tasks.py
```diff
@@ -73,6 +73,14 @@ class TaskBase(BaseModel):
         default=None,
         description=PROXY_LOCATION_DOC_STRING,
     )
+
+    @field_validator("proxy_location")
+    @classmethod
+    def validate_proxy_location(cls, v: ProxyLocationInput) -> ProxyLocationInput:
+        if isinstance(v, str) and not isinstance(v, ProxyLocation):
+            return validate_proxy_location_str(v)
+        return v
```
AronPerez: This exists in `TaskBase` (schemas/tasks.py) and `CreateBrowserSessionRequest` (schemas/browser_sessions.py). Can you extract this to a shared validator or create a custom Pydantic type please?

AronPerez (changes requested):

skyvern/schemas/runs.py
navanpreet: Done. Consolidated into a single `is_valid_proxy_url()` in `schemas/runs.py`. `validate_proxy_location_str()` now delegates to it, and `browser_factory.py` imports it from there instead of defining its own copy. Also created a `ValidatedProxyLocationInput` annotated type so the duplicated `@field_validator` in TaskBase and `CreateBrowserSessionRequest` is gone too.

skyvern/forge/sdk/schemas/tasks.py
navanpreet: Done. Created a `ValidatedProxyLocationInput` annotated type in `schemas/runs.py` that uses a `BeforeValidator` to handle the string proxy URL check. Both `TaskBase` and `CreateBrowserSessionRequest` now use it as the field type directly, so the duplicated `@field_validator` is removed from both.

AronPerez (approved):

AronPerez:
Approved pending merge conflict resolved

PR #4976, opened 2026-03-04
fix: fail immediately on non-retriable navigation errors (SKY-8875)

andrewneilson:
## Summary

  - Classify navigation failures into three buckets:
    - permanent: `ERR_INVALID_URL`
    - proxy-sensitive: `ERR_NAME_NOT_RESOLVED`, `ERR_NAME_RESOLUTION_FAILED`, `ERR_CERT_*`, `ERR_SSL_*`
    - transient: other `net::ERR` failures
  - Skip same-context retries for permanent and proxy-sensitive errors in `navigate_with_retry()`
  - Preserve the outer context-recreation retry in `get_or_create_page()` for proxy-sensitive and transient
  errors, since recreating the browser context may recover by selecting a different proxy node
  - Only permanent failures like `ERR_INVALID_URL` now bypass both retry layers
  - Use `FailedToNavigateToUrl.error_message` in the outer retry path when available instead of relying on
  `str(e)`

  ## Test plan

  - [x] `ruff check skyvern/constants.py skyvern/webeye/navigation.py skyvern/webeye/real_browser_state.py
  tests/unit_tests/webeye/test_navigation_retry.py`
  - [x] `ruff format --check skyvern/constants.py skyvern/webeye/navigation.py skyvern/webeye/
  real_browser_state.py tests/unit_tests/webeye/test_navigation_retry.py`
  - [x] Unit coverage for:
    - inner retry short-circuit for permanent and proxy-sensitive errors
    - inner retry exhaustion for transient errors
    - transient recovery on retry
    - outer retry skip for permanent errors
    - outer context-recreation retry for DNS errors
    - outer context-recreation retry for transient `net::ERR` errors

### DIFF
--- a/skyvern/webeye/real_browser_state.py
+++ b/skyvern/webeye/real_browser_state.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import asyncio
 import random
 import time
+from typing import Literal
 from urllib.parse import urlparse
 
 import structlog
@@ -21,6 +22,7 @@ from skyvern.schemas.runs import ProxyLocationInput
 from skyvern.webeye.browser_artifacts import BrowserArtifacts, VideoArtifact
 from skyvern.webeye.browser_factory import BrowserCleanupFunc, BrowserContextFactory
 from skyvern.webeye.browser_state import BrowserState
+from skyvern.webeye.navigation import is_permanent_navigation_error, navigate_with_retry
 from skyvern.webeye.scraper import scraper
 from skyvern.webeye.scraper.scraped_page import CleanupElementTreeFunc, ScrapedPage, ScrapeExcludeFunc
 from skyvern.webeye.utils.page import ScreenshotMode, SkyvernFrame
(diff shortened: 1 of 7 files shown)


AronPerez (commented):
## Auto-Review: PR #4976

### Build Status
- All CI checks pass.

### Findings Summary
- 1 bug (see inline comment) — outer retry loop undermines the non-retriable logic
- 1 suggestion (see inline comment) — tests duplicate production logic
- Good idea with clean implementation, but an existing outer retry loop will still retry non-retriable errors.

### Agreement with Claude's Review
- No Claude review found.

---
*Reviewed by AronPerez's auto-reviewer*

skyvern/webeye/real_browser_state.py
```diff
@@ -139,11 +144,21 @@ async def navigate_to_url(self, page: Page, url: str, retry_times: int = NAVIGAT
                     return
 
                 except Exception as e:
+                    error_str = str(e)
+
+                    if any(pattern in error_str for pattern in NON_RETRIABLE_NAV_ERRORS):
+                        LOG.warning(
```
AronPerez: **Bug**: The outer retry in `get_or_create_page` (further down in this file) catches any exception whose `str()` contains `"net::ERR"` and retries by closing the browser context. Since `FailedToNavigateToUrl` embeds the original error in its message (e.g., `"Failed to navigate to url ... Error message: net::ERR_NAME_NOT_RESOLVED"`), these non-retriable errors will still be caught and retried at the `get_or_create_page` level — partially defeating the purpose of this PR. Consider either:
1. Having `get_or_create_page` re-raise `FailedToNavigateToUrl` without retrying, or
2. Adding the non-retriable check there as well.

tests/unit_tests/webeye/test_navigation_retry.py
```diff
@@ -0,0 +1,84 @@
...
+
+import asyncio
+from unittest.mock import AsyncMock
+
+import pytest
+
+from skyvern.constants import NON_RETRIABLE_NAV_ERRORS
+from skyvern.exceptions import FailedToNavigateToUrl
+
+
```
AronPerez: **Suggestion**: These tests re-implement `navigate_to_url`'s retry logic in a local helper rather than testing the real `RealBrowserState.navigate_to_url`. If the production code changes its retry behavior, these tests will still pass. Consider mocking the heavy dependencies and testing the real method, or extracting the retry logic into a standalone testable function.

Also worth considering: adding `net::ERR_SSL_` to the non-retriable patterns in `constants.py` — errors like `ERR_SSL_PROTOCOL_ERROR` indicate equally non-retriable TLS incompatibilities.

andrewneilson:
made these changes to address feedback:

```
    fix: three-tier nav error classification for proxy-aware retry

    Split navigation errors into permanent (ERR_INVALID_URL) and
    proxy-sensitive (DNS/TLS) categories. Permanent errors bypass all
    retry layers. Proxy-sensitive errors skip same-context retry but
    still get the outer context-recreation path in get_or_create_page,
    which can pick a different proxy node.

    Also extracts navigate_with_retry into a shared helper and uses
    the typed FailedToNavigateToUrl.error_message field in the outer
    except block instead of falling back to str(e).
```

skyvern/webeye/real_browser_state.py
andrewneilson: Thanks for catching that. I ended up not disabling the outer retry for all of the error types in question since the two layers serve different purposes:

- `navigate_with_retry()` retries in the same page/context/proxy, which is not useful for `ERR_NAME_*`,
`ERR_CERT_*`, and `ERR_SSL_*`
- the outer `get_or_create_page()` retry recreates the browser context, which is the only path that can pick a different proxy node. DNS/TLS failures can be proxy-specific so fully skipping the outer retry would turn recoverable proxy failures into immediate task failures.
- I've kept permanent errors (currently just `ERR_INVALID_URL`) in a separate category that skips both

wintonzheng (approved):
Looks really good

LawyZheng (approved):

wintonzheng (approved):
## Review: LGTM ✅

Well-structured PR. Error classification logic is correct, extraction into `navigation.py` improves testability, and the changes are backwards-compatible.

### Verified

- **No regression risk** — `navigate_to_url` contract is unchanged (same signature, same exception type). Only behavioral change is faster failure for `ERR_INVALID_URL` (skips outer proxy retry), which is correct.
- **Error buckets are sound** — permanent/proxy-sensitive/transient classification matches Chromium error semantics.
- **Good improvements** — `raise e` → `raise` (preserves traceback), `time.time()` → `time.monotonic()` (clock-immune), `ast.Str` cleanup (forward-compat with 3.14).

### One consideration (non-blocking)

`ERR_ABORTED` is currently unclassified — it can indicate intentional page stops or proxy failures. Worth considering whether it should be bucketed in a follow-up.

### Reminder

Confirm `tests/unit/test_real_browser_state_navigate.py` (not in the diff) still passes — the functional contract is preserved so they should, but good to verify.

PR #4989, opened 2026-03-05
feat: workflow copilot agent using OpenAI Agents SDK

andrewneilson:
> [!IMPORTANT]
> I'm posting this for feedback, but note that this still needs a lot of iteration to satisfy the use cases we have in mind.

## Summary

- Adds a workflow copilot agent powered by the OpenAI Agents SDK with LiteLLM for multi-provider LLM support
- Implements multi-turn tool-use agent for building workflows via chat, with structured context tracking, loop detection, and streaming SSE responses
- Includes new copilot modules: agent, context, enforcement, hooks, loop detection, MCP adapter, model resolver, output utils, runtime, streaming adapter, and tools
- Refactors workflow copilot routes to use the new agent architecture
- Adds frontend chat UI improvements for the copilot experience
- Adds unit tests for copilot sessions, SDK contracts, route handling, and output utils

## Test plan
- [ ] Run `pytest tests/unit/` and `pytest tests/unit_tests/forge/sdk/copilot/` to verify new tests pass
- [ ] Start the service and test the copilot chat flow end-to-end via the UI
- [ ] Verify streaming SSE responses work correctly in the browser
- [ ] Test multi-turn conversations with workflow creation and modification

### DIFF
--- a/skyvern/forge/prompts/skyvern/workflow-copilot-agent.j2
+++ b/skyvern/forge/prompts/skyvern/workflow-copilot-agent.j2
@@ -1,0 +1,204 @@
+You are an expert Skyvern Workflow assistant with access to tools that let you build, validate, test, and debug browser automation workflows.
+
+WORKFLOW KNOWLEDGE BASE:
+
+{{ workflow_knowledge_base }}
+
+CURRENT WORKFLOW YAML:
+
+{% if workflow_yaml %}
+The user's current workflow definition is:
+
+```yaml
+{{ workflow_yaml }}
+```
+{% else %}
+The user is starting with an empty workflow.
+{% endif %}
+
+PREVIOUS CONTEXT:
+
+{% if chat_history %}
+Recent conversation history:
+{{ chat_history }}
+{% endif %}
+
+{% if global_llm_context %}
+STRUCTURED MEMORY (long-term context — preserve and extend each turn):
+```json
+{{ global_llm_context }}
+```
+{% endif %}
+
+{% if not chat_history and not global_llm_context %}
+No previous context available.
+{% endif %}
+
+DEBUGGER RUN INFORMATION:
+
+{% if debug_run_info %}
+The user has run the workflow in the debugger. Here's the most recent block execution information:
+
+{{ debug_run_info }}
+
+Use this information to help diagnose issues, suggest fixes, or explain what might be going wrong.
+{% else %}
+No debugger run information available.
+{% endif %}
+
+TOOL USAGE INSTRUCTIONS:
+
+You have tools available in this run. Use them to build better workflows:
+{{ tool_usage_guide }}
+
+IMPORTANT: NEVER call the same tool more than twice in a row. If a tool succeeded, move on to the NEXT step — do not repeat the same call. Each tool call should advance your progress toward the goal.
+
+WORKFLOW-FIRST EXECUTION PATH (INCREMENTAL):
+Default path for build/debug requests:
+1. Do NOT draft the full workflow upfront.
+2. Start by adding only the next navigation block needed to inspect or advance the target website from the current state.
+3. Call update_workflow with known-good blocks plus that next block.
+4. Call run_blocks_and_collect_debug on ALL current block labels.
+5. Use run evidence (status, failure_reason, output, visible_elements_html, screenshot) to decide whether the task still looks fixable.
+6. If fixable, keep known-good blocks and add/refine only the next block, then repeat.
+7. If not fixable, stop and explain the specific evidence to the user.
+
+IMPORTANT: This incremental loop is INTERNAL. Do not stop after the first successful block run.
+Continue iterating until one of these is true:
+1. You have a completed workflow proposal that covers the user's full requested outcome and the latest test run is successful.
+2. You have strong evidence from the latest run that the request is infeasible on the target site.
+3. You are blocked by missing required user input (for example: missing credential, unresolved ambiguity, or required human-only value).
+
+DIRECT BROWSER TOOLS (OPTIONAL DIAGNOSTICS):
+Prefer workflow-run evidence first. Use navigate_browser, get_browser_screenshot, evaluate, click, and type_text only when:
+1. The user explicitly asks to inspect or interact with the live page.
+2. The latest workflow run evidence is missing, ambiguous, or contradictory.
+3. You need to validate a specific UI hypothesis before changing the workflow.
+
+CRITICAL RULES WHEN USING DIRECT BROWSER TOOLS:
+- If you call navigate_browser, you MUST call at least one observation tool (evaluate, get_browser_screenshot, click, or type_text) before responding.
+- NEVER answer questions about page content, link counts, link targets, or page structure from memory or training data. Use evaluate or get_browser_screenshot.
+- Use evaluate for reading/inspection; do NOT use evaluate to click or type. Use click and type_text for interactions.
+- CSS SELECTORS ONLY: The click and type_text tools use standard CSS selectors via Playwright. jQuery pseudo-selectors (:contains(), :eq(), :first, :visible, :has()) are NOT valid CSS and will fail. Use standard selectors like 'button.classname', 'a[href*="keyword"]', '#id', 'tr:nth-of-type(2) td a'. To select by text content, use evaluate with JavaScript: `[...document.querySelectorAll('a')].find(el => el.textContent.includes('Download'))`.
+
+FEASIBILITY CHECK (POST-UPDATE, RESULT-DRIVEN):
+Assess feasibility after each incremental update using the latest run:
+1. Evaluate the newest run_blocks_and_collect_debug result only.
+2. If the result suggests a prompt/config issue (wrong goal wording, missing credential, bad ordering, timeout, extraction mismatch), treat as FIXABLE and continue.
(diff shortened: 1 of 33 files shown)


skyvern/forge/prompts/skyvern/workflow-copilot-agent.j2
```diff
@@ -0,0 +1,203 @@
...
+{{ workflow_yaml }}
+```
+{% else %}
+The user is starting with an empty workflow.
+{% endif %}
+
+PREVIOUS CONTEXT:
+
+{% if chat_history %}
+Recent conversation history:
```
suchintan, line 22: (no action needed), I wonder if we will need to compact this in the future

skyvern/forge/sdk/copilot/output_utils.py
```diff
@@ -0,0 +1,92 @@
...
+    """Strip large/binary fields from tool results before sending to the LLM."""
+    sanitized = dict(result)
+    for key in ("action", "browser_context", "artifacts", "timing_ms"):
+        sanitized.pop(key, None)
+
+    data = sanitized.get("data")
+    if isinstance(data, dict):
+        data = dict(data)
+        if "screenshot_base64" in data:
+            data["screenshot_base64"] = "[base64 image omitted — screenshot was taken successfully]"
```
marcmuon, line 43: does this mean the agent can't visually diagnose failures in the debug loop? If true, I'm wondering if passing the screenshots thought would improve debug loop

skyvern/forge/sdk/copilot/enforcement.py
```diff
@@ -0,0 +1,185 @@
...
+                "Enforcement nudge",
+                nudge_type="post_update",
+                nudge_count=ctx.post_update_nudge_count,
+            )
+        elif nudge == POST_NAVIGATE_NUDGE:
+            LOG.info("Enforcement nudge", nudge_type="post_navigate", nudge_count=None)
+        else:
+            LOG.info("Enforcement nudge", nudge_type="intermediate_success", nudge_count=None)
+
+        current_input = result.to_input_list() + [{"role": "user", "content": nudge}]
```
marcmuon: is context accumulation going to be a problem here? since we have max token limits. `to_input_list` looks like it will carry the full history of all prior iterations https://openai.github.io/openai-agents-python/ref/result/#agents.result.RunResultBase.to_input_list

skyvern/forge/sdk/copilot/output_utils.py
```diff
@@ -0,0 +1,92 @@
...
+        if "visible_elements_html" in data and data["visible_elements_html"]:
+            html = data["visible_elements_html"]
+            if len(html) > 3000:
+                data["visible_elements_html"] = html[:3000] + "\n... [truncated]"
```
marcmuon, line 47: just curious how this 3000 is set and if it could be too small for complex pages

skyvern/forge/sdk/copilot/tools.py
```diff
@@ -0,0 +1,757 @@
...
+        block_result: dict[str, Any] = {
+            "label": block.label,
+            "block_type": block.block_type.name if hasattr(block.block_type, "name") else str(block.block_type),
+            "status": block.status,
+        }
+        if block.failure_reason:
+            block_result["failure_reason"] = block.failure_reason
+        if hasattr(block, "output") and block.output:
+            block_result["extracted_data"] = block.output
+        results.append(block_result)
```
marcmuon, line 405: i wonder if we appended the actual actions taken from our actions table and extracted a small trace if it would help the agent in the debug loop by seeing the failure reason. i'm not totally sure

something like 

```
for block in blocks:
      if not block.task_id:
          continue
      actions = await app.DATABASE.get_tasks_actions(
          task_id=block.task_id,
          organization_id=ctx.organization_id,
      )
      block_result["action_trace"] = [
          {
              "action": a.action_type,
              "status": a.status,
              "reasoning": a.reasoning[:150] if a.reasoning else None,
              "element": a.element_id,
          }
          for a in actions[:20] 
      ]
```

skyvern/forge/sdk/copilot/output_utils.py
andrewneilson: I'll take another look at this tonight. It's a result of me paring back to make sure we're relying on skyvern MCP as much as possible.

skyvern/forge/sdk/copilot/output_utils.py
andrewneilson: It's a (somewhat arbitrary) safeguard to keep the debug output bounded. Since you mentioned it, I realized there is a redundant limit in tools.py to fix.

The intent is to just give the model a rough hint of what's on the page from the scraped visible elements tree rather than a precise dom reference. I haven't run into it in my limited testing but I wouldn't be surprised if we need to revisit it.

skyvern/forge/sdk/copilot/enforcement.py
andrewneilson: good callout. I am interested in seeing if I can use https://openai.github.io/openai-agents-python/sessions/ to better manage what we pull into `current_input`. For example we could keep the last N turns, or last N turns plus a compact summary.

skyvern/forge/sdk/copilot/tools.py
andrewneilson: I need to do a lot more testing but I agree this could be a good signal

andrewneilson:
```
   feat: copilot agent vision, context management, and enforcement improvements

    Add screenshot vision pipeline (resize/compress to JPEG, inject as synthetic
    user messages), multi-layered context window overflow prevention with graduated
    pruning, new enforcement rules for workflow-first exploration and failed-test
    iteration, action traces for failed blocks, logfire tracing integration,
    SSE connection stability fix, jQuery selector guard, and streaming disconnect
    handling.

    Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
```

skyvern/forge/sdk/db/agent_db.py
```diff
@@ -630,6 +630,34 @@ async def get_tasks_actions(self, task_ids: list[str], organization_id: str | No
...
             raise
 
+    async def get_recent_actions_for_tasks(
+        self,
+        task_ids: list[str],
+        organization_id: str | None = None,
+        per_task_limit: int = 15,
+    ) -> list[ActionModel]:
+        """Return the most recent N actions per task as raw models (no hydration)."""
+        try:
```
wintonzheng, line 640: **N+1 query**: This iterates `task_ids` and issues a separate `SELECT` per task ID within the same session. For workflows with many blocks, this creates N separate DB round-trips.

Consider using a single query with `ActionModel.task_id.in_(task_ids)` combined with a window function (e.g. `ROW_NUMBER() OVER (PARTITION BY task_id ORDER BY created_at DESC)`) or application-level grouping to enforce the per-task limit in one round-trip.

skyvern/forge/sdk/copilot/enforcement.py
```diff
@@ -0,0 +1,485 @@
...
+def is_synthetic_user_message(item: Any) -> bool:
+    """Return True if item is a screenshot or nudge (not a real user turn)."""
+    return is_screenshot_message(item) or _is_nudge_message(item)
+
+
+def _prune_input_list(items: list[Any]) -> list[Any]:
+    """Remove old screenshot messages and truncate large tool outputs.
+
+    Keeps only the most recent synthetic screenshot message.
+    Replaces older ones with a short text placeholder.
```
wintonzheng, line 267: **Token estimation is very rough**: `len(json.dumps(item)) / 4` is a coarse approximation that doesn't account for JSON encoding overhead (keys, quotes, escapes inflate char count) or model-specific tokenization differences.

There's already a `skyvern/utils/token_counter.py` that does proper tokenization via tiktoken. Consider reusing that instead of rolling a separate estimator here. If performance is a concern (tiktoken can be slow on large inputs), at minimum consider calibrating `CHARS_PER_TOKEN` against actual token counts — the current value of 4 is likely too generous for JSON-heavy content where keys and structural characters dominate.

If the estimate is too low, the 90k budget gets exceeded → context window errors → `aggressive_prune`. If too high, useful context is dropped prematurely.

skyvern/forge/sdk/copilot/enforcement.py
```diff
@@ -0,0 +1,485 @@
...
+        result.append(item)
+    return result
+
+
+def estimate_tokens(items: list[Any]) -> int:
+    """Rough token estimate for an input list. Errs on the high side."""
+    total = 0
+    for item in items:
+        if isinstance(item, dict):
+            total += _estimate_dict_tokens(item)
```
wintonzheng, line 301: **`aggressive_prune` can destroy critical context**: When triggered (either by context window errors or the session factory's layer-4 fallback), this keeps only the original user message + the last 7 non-screenshot items. All intermediate assistant reasoning, tool calls, and tool results are dropped.

If the agent's understanding of what to do next was encoded in those intermediate messages (e.g., a multi-step plan it was executing), it will "forget" its plan and potentially repeat or contradict earlier actions.

Consider:
1. Preserving the last assistant message that contained the agent's plan/reasoning, not just raw tool pairs
2. Adding a synthetic summary message that captures key state (e.g., "You have already completed steps A, B, C. The current workflow has N blocks. Next step is D.") before pruning
3. Logging/metrics when aggressive_prune fires so you can monitor how often this recovery path is hit in production

andrewneilson:
closing this in favour of a series of PRs against the cloud. repo

PR #5373, opened 2026-04-03
fix: prevent S3 token expiry on Hetzner by adding TTL + retry to AWS client (SKY-8743)

AronPerez:


wintonzheng:
stale PR. closing. the diff was already merged by another sync PR

PR #5513, opened 2026-04-15
Add Yutori Navigator as a CUA engine

lawrencechen98:
## Summary

Adds [Yutori Navigator](https://docs.yutori.com/reference/navigator) as a new CUA engine (`engine: "yutori-navigator"`) using the [Yutori Python SDK](https://github.com/yutori-ai/yutori-sdk-python). Follows the same architecture pattern as the UI-TARS integration.

### What's included

- **`YutoriNavigatorLLMCaller`** extending `LLMCaller` — artifact persistence, cost/token tracking, and error handling via the base class
- **API routing** through `_dispatch_llm_call()` with `_call_yutori_navigator()`, converting responses to `litellm.ModelResponse` for standard stats tracking
- **Response parser** mapping Navigator's tool_calls to Skyvern actions, using the SDK for coordinate denormalization and key mapping
- **Expanded tool support** — `extract_elements`, `find`, `set_element_value`, `execute_js` mapped to a new generic `ExecuteJsAction` that goes through Skyvern's action pipeline
- **Ref-based targeting** — Navigator can target elements by ref ID; resolved inline to coordinates via the SDK's `get_element_by_ref.js`
- **Stop-and-summarize** — on max steps, Navigator produces a progress summary instead of failing silently
- **Unified tool result handling** — all action results (browser + JS) flow back to Navigator as tool responses with actual execution data
- **New action types**: `GoForwardAction` (`page.go_forward()`), `ExecuteJsAction` (`page.evaluate()`), `ReloadPageAction` handler

### Engine plumbing

`RunType.yutori_navigator` / `RunEngine.yutori_navigator` wired through all standard paths: `schemas/runs.py`, `run_service.py`, `task_v1_service.py`, `agent_protocol.py`, `background_task_executor.py`, `client/types`, frontend types + engine selector, OpenAPI schemas, SDK docs.

### Configuration

```
ENABLE_YUTORI=true
YUTORI_API_KEY=<key>
YUTORI_MODEL=n1.5-latest
YUTORI_TOOL_SET=                          # empty = core, or "browser_tools_expanded-20260403"
BROWSER_WIDTH=1280
BROWSER_HEIGHT=800
```

### Dependencies

- `yutori>=0.5.0` (Yutori Python SDK) — wraps `openai` internally (already a Skyvern dependency)
- `httpx[socks]>=0.26.0` override to resolve fastmcp/yutori httpx version conflict

## Test plan

- [x] `uv sync` installs cleanly
- [x] Submit task with `engine: "yutori-navigator"` — completes successfully
- [x] LLM request/response artifacts visible in Skyvern UI Diagnostics tab
- [x] Token counts populated on steps in DB
- [x] Expanded tool set tasks execute JS and return results
- [x] Max steps triggers stop-and-summarize with progress output
- [x] 15-website WebVoyager CUA benchmark: 12/15 completed (3 infra timeouts from concurrent load)


YUTORI_INTEGRATION.md
wintonzheng, line 1: Can we remove this .md file? we don't keep development / design doc like this

skyvern/webeye/actions/actions.py
```diff
@@ -290,6 +296,10 @@ def __repr__(self) -> str:
 
 class NullAction(Action):
     action_type: ActionType = ActionType.NULL_ACTION
+    # Some integrations use NullAction as a successful no-op placeholder that still
+    # needs to wait or surface a tool-result string back to the model.
+    sleep_seconds: float = 0
+    result_data: dict[str, Any] | list | str | None = None
```
wintonzheng: to keep the schema / naming consistent with other places where we have data output like this, let's use `output`:
```suggestion
    output: dict[str, Any] | list | str | None = None
```

skyvern/forge/sdk/api/llm/yutori_navigator_response.py
```diff
@@ -0,0 +1,310 @@
...
+
+    # ---- Completion / Wait ----
+    if action_type == YutoriNavigatorActionType.STOP:
+        summary = args.get("summary") or args.get("message")
+        return CompleteAction(data_extraction_goal=summary, **bp)
+
+    if action_type in (YutoriNavigatorActionType.WAIT, YutoriNavigatorActionType.SLEEP):
+        duration_value = args.get("duration")
+        duration = max(0.0, min(float(5 if duration_value is None else duration_value), 100.0))
+        return NullAction(
```
wintonzheng: We already have the `WaitAction`. Let's use that

skyvern/forge/sdk/api/llm/yutori_navigator_response.py
```diff
@@ -0,0 +1,310 @@
...
+            sleep_seconds=duration,
+            result_data=f"Waited {duration:g}s",
+            **bp,
+        )
+
+    # Legacy extract action
+    if action_type == YutoriNavigatorActionType.EXTRACT_CONTENT:
+        summary = args.get("summary") or args.get("query") or args.get("goal") or "Content extracted"
+        return CompleteAction(data_extraction_goal=summary, **bp)
```
wintonzheng: can we add an `output` field to the CompleteAction?


```suggestion
        return CompleteAction(output=summary, **bp)
```

wintonzheng (changes requested):
Looks pretty good! No major blocker. Left small comments. Let's address those and resolve the conflict.

skyvern/forge/sdk/api/llm/yutori_navigator_response.py
lawrencechen98: We were unsure if it was a problem to use WaitAction, given it's implementation has:
`return [ActionFailure(exception=Exception("Wait action is treated as a failure"))]`

Our model uses WaitAction as a normal tool, so wasn't sure if this was the right approach. We would also ideally want the WaitAction to return a result confirming a positive wait behavior, e.g. "Waited for 1s". Which is why we opted to reframe our "wait" action as a `NullAction` as it felt more similar at first glance.

Could we change the WaitAction to allow overrides for action success behavior instead? Happy to switch over if this is not an issue, or let us know if you have other suggestions. Thank you!

lawrencechen98: Update: we realized we can leverage WaitAction as is, but just simply override the tool response shown to the model. So the ActionFailure mechanism is intact, while the model still sees expected tool results.

wintonzheng (approved):

lawrencechen98:
@wintonzheng Can you help us approve the remaining workflows that are awaiting approval? Thank you!

PR #5529, opened 2026-04-16
feat(SKY-8879) copilot-stack/11: agent orchestrator + prompt

andrewneilson:


andrewneilson:
The 6000+ lines is because tests from previous PRs got moved to `tests/unit` from `tests/unit_tests`

This was fixed in another cloud-side PR that didn't get picked up by the OSS sync because it incorrectly failed the "OSS relevant changes" check. So we do want these pulled in, but they weren't part of the original `copilot-stack/11` PR.

For the CodeQL warning: I will address on the cloud side but I don't think it should block merge.

wintonzheng (approved):

PR #5531, opened 2026-04-17
feat(SKY-8879) copilot-stack/12: wire-up (flag + dispatch + frontend)

andrewneilson:


andrewneilson:
ci failure is because there's a workflow change this depends on that doesn't get synced from cloud. Manually adding it to this pr

PR #5537, opened 2026-04-17
chore: persist frontend package lock update for Skyvern UI

minhazjisan85:


skyvern-frontend/package-lock.json
AronPerez, line 9461: Question, what is this trying to achieve

PR #5630, opened 2026-04-23
Hide requested API reference pages from Mintlify

marcmuon:
## Summary
- mark only POST /api/v1/billing/checkout and POST /v1/sdk/run_action as x-excluded in the Mintlify docs OpenAPI file
- add x-excluded to the SDK action FastAPI OpenAPI metadata so future API syncs preserve the docs exclusion without removing SDK support

## Verification
- jq empty docs/api-reference/openapi.json
- checked the two target operations return x-excluded=true and the other billing operations remain unset
- python3 -m py_compile skyvern/forge/sdk/routes/sdk.py
- git diff --check -- docs/api-reference/openapi.json skyvern/forge/sdk/routes/sdk.py
- pre-commit hooks passed during commit

## Notes
- This intentionally hides only these two public docs pages:
  - /docs/api-reference/create-checkout-session
  - /docs/api-reference/sdk/run-an-sdk-action
- Live URLs will remain 200 until this merges and Mintlify redeploys.

### DIFF
--- a/docs/api-reference/openapi.json
+++ b/docs/api-reference/openapi.json
@@ -2168,6 +2168,7 @@
     },
     "/v1/sdk/run_action": {
       "post": {
+        "x-excluded": true,
         "tags": [
           "SDK"
         ],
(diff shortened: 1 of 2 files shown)


wintonzheng (commented):
shall we also exclude it from our openapi.json?

can be don with the "include_in_schema=False" flag in the routes

wintonzheng (approved):

marcmuon:
> shall we also exclude it from our openapi.json?
> 
> can be don with the "include_in_schema=False" flag in the routes

doesnt that remove it from generated clients? im just trying to hide it from the docs with the mintifly arg with this. the billing one should have already been hidden in fern, but it needs the mintifly one to hide it in docs now for example

PR #5705, opened 2026-04-28
security: mask secrets in LLM prompt artifacts

johnpippett:
## Summary

Fixes the secret-artifact portion of the 2026-04-28 security audit finding by preventing workflow secrets and runtime sensitive values from being persisted in LLM prompt artifacts.

## Changes

- Moved LLM prompt masking to the artifact persistence boundary in ArtifactManager.prepare_llm_artifact.
- Mask workflow secrets through WorkflowRunContext.mask_secrets_in_data before LLM prompt artifacts are created.
- Mask runtime sensitive values from SkyvernContext.sensitive_values, including TOTP verification codes and TOTP secrets.
- Fail closed if prompt masking fails by persisting a redacted placeholder instead of raw prompt text.
- Apply the same masking path to bundled step-archive LLM prompt entries.

## Out of scope

Prompt-injection/template-boundary hardening is intentionally removed from this PR per review feedback.

## Test results

- uv run pytest tests/unit/test_bulk_artifact_creation.py tests/unit/test_secret_credentials.py -q -- 12 passed
- uv run ruff check skyvern/forge/agent.py skyvern/forge/sdk/artifact/manager.py tests/unit/test_bulk_artifact_creation.py -- passed

### DIFF
--- a/skyvern/forge/sdk/artifact/manager.py
+++ b/skyvern/forge/sdk/artifact/manager.py
@@ -108,6 +108,30 @@ class ArtifactManager:
         self._step_archives: dict[str, StepArchiveAccumulator] = {}
 
     @staticmethod
+    def _mask_llm_prompt_artifact_data(data: bytes, artifact_type: ArtifactType) -> bytes:
+        if artifact_type != ArtifactType.LLM_PROMPT:
+            return data
+
+        try:
+            prompt = data.decode("utf-8")
+            context = skyvern_context.current()
+            if context and context.workflow_run_id:
+                workflow_run_context = app.WORKFLOW_CONTEXT_MANAGER.get_workflow_run_context(
+                    context.workflow_run_id
+                )
+                prompt = workflow_run_context.mask_secrets_in_data(prompt)
+
+            if context and context.sensitive_values:
+                for sensitive in sorted(context.sensitive_values, key=len, reverse=True):
+                    if sensitive:
+                        prompt = prompt.replace(sensitive, "*****")
+
+            return prompt.encode("utf-8")
+        except Exception:
+            LOG.warning("Failed to mask LLM prompt artifact; redacting prompt artifact")
+            return b"[LLM prompt artifact redacted: secret masking failed]"
+
+    @staticmethod
     def _build_artifact_model(
         artifact_id: str,
         artifact_type: ArtifactType,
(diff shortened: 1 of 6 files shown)


andrewneilson (changes requested):
@johnpippett Thank you for raising this. Sorry for the delay in a response.

There are two threads tangled together in this diff and they live at different layers:

1. **artifact masking**: this is the part I think we can get via this PR. 
   - **Could you scope the PR down to just this**: it belongs in [`ARTIFACT_MANAGER.prepare_llm_artifact`](https://github.com/Skyvern-AI/skyvern/blob/80b6f67edec41c763a2fd60a8688f41b300a0264/skyvern/forge/sdk/artifact/manager.py#L970) rather than the LLM handler, it should fail closed, and the TOTP-code lifecycle / set ordering parts need a second pass. Let me know if this is not clear.
2. **prompt injection**: we already have [`escape_code_fences` in `skyvern/utils/strings.py`](https://github.com/Skyvern-AI/skyvern/blob/80b6f67edec41c763a2fd60a8688f41b300a0264/skyvern/utils/strings.py#L52) doing similar work, and the right fix for this is a template level boundary rather than per-call-site fixes. This is easier to drive internally than to try to resolve async here.

johnpippett:
Scoped this down to artifact masking only per review.

Updated in d3262414:
- removed the prompt-injection/template sanitization changes from the PR diff
- moved LLM prompt masking into ArtifactManager.prepare_llm_artifact
- added fail-closed behavior so raw prompt text is not persisted if masking fails
- registered TOTP verification codes/secrets as runtime sensitive values
- covered bundled step-archive LLM prompt entries too

Verification:
- uv run pytest tests/unit/test_bulk_artifact_creation.py tests/unit/test_secret_credentials.py -q
- uv run ruff check skyvern/forge/agent.py skyvern/forge/sdk/artifact/manager.py tests/unit/test_bulk_artifact_creation.py

andrewneilson:
@johnpippett Thanks, this is much closer to what I was asking for. The other work landed in #6372.

I think there is one gap left before this is ready: `TextPromptBlock` still saves an `LLM_PROMPT` artifact through a workflow-run-block bulk artifact path, so it does not hit the new masking helper. Could you pull that path under the same masking/fail-closed behaviour too?

skyvern/forge/sdk/artifact/manager.py
andrewneilson, line 1: This path still needs to mask `ArtifactType.LLM_PROMPT` data before upload. `TextPromptBlock` builds a rendered LLM prompt, adds it as an `LLM_PROMPT`, and persists it through `create_workflow_run_block_artifacts`, so it bypasses the new `prepare_llm_artifact` masking.

Can we run `LLM_PROMPT` bytes through the same helper here as well? My preference is to make the artifact manager enforce this wherever it accepts prompt artifacts, rather than relying on each caller to remember the right path.

PR #5719, opened 2026-04-29
fix: quickstart fails on Windows — missing DATABASE_STRING

suchintan:
## Summary
- `setup_postgresql()` created a Docker Postgres container but never wrote `DATABASE_STRING` to `.env`, so `migrate_db()` fell back to the SQLite default — which failed on Windows because `~/.skyvern/` didn't exist yet ("Unable to open the database file")
- `migrate_db()` now reloads `Settings()` so env vars set by `setup_postgresql()` are picked up, and calls `_ensure_sqlite_dir()` as a safety net for the SQLite fallback path
- Docker-not-found error now distinguishes "not installed" (with install link + `--database-string` workaround) vs "not running" (tells user to start Docker Desktop)

## Test plan
- [ ] Run `skyvern quickstart` on Windows with Docker running — should migrate against Postgres, not SQLite
- [ ] Run `skyvern quickstart` on Windows without Docker installed — should show actionable error with install link
- [ ] Run `skyvern quickstart` on Windows with Docker installed but stopped — should show "start Docker Desktop" message
- [ ] Run `skyvern quickstart` on macOS/Linux — existing behavior unchanged


marcmuon (changes requested):
Reviewed end-to-end. The core Windows fix is good — calling `update_or_add_env_var("DATABASE_STRING", ...)` after both setup paths in `setup_postgresql()`, plus reloading `Settings()` and calling `_ensure_sqlite_dir()` inside `migrate_db()`, correctly closes the gap where Postgres was provisioned but Alembic still ran against the SQLite default.

Two blocking items before merge:

**1. `host.docker.internal` written to `.env` will silently break plain Linux Docker Engine**

`skyvern/cli/quickstart.py:169` writes `BROWSER_REMOTE_DEBUGGING_URL=http://host.docker.internal:9222/` to `.env` for the "control your own browser" path, but the `skyvern` service in `docker-compose.yml` does not declare `extra_hosts: ["host.docker.internal:host-gateway"]`. This works on Docker Desktop (mac/Windows) but silently fails on plain Linux Docker Engine — the most common OSS self-host configuration. The PR's stated goal is to improve cross-platform quickstart, so baking a Docker-Desktop-only URL into self-hosted users' `.env` is a regression for Linux.

Fix: either add `extra_hosts: ["host.docker.internal:host-gateway"]` to the `skyvern` service in `docker-compose.yml` (works on Desktop and Engine), or branch on platform / print a Linux-specific note.

**2. Stale `DATABASE_STRING` default in `update_or_add_env_var` seed dict**

`skyvern/cli/llm_setup.py:45` seeds a fresh `.env` with `DATABASE_STRING=postgresql+psycopg://skyvern@localhost/skyvern` (no `:5432` port), which conflicts with the new `DOCKER_POSTGRES_DATABASE_STRING = "postgresql+psycopg://skyvern@localhost:5432/skyvern"` constant in `database.py`. This is the same bug class this PR is fixing — two slightly different "default DATABASE_STRING" values that will bite when one is updated and the other isn't. Please import and reuse `DOCKER_POSTGRES_DATABASE_STRING` here. (`skyvern/cli/database.py:128` also still uses the literal — line 253 already uses the constant.)

Once these two are addressed I'm happy to approve.

suchintan:
Both items addressed in the latest push:

1. Added `extra_hosts: ["host.docker.internal:host-gateway"]` to the `skyvern` service in docker-compose.yml — works on both Docker Desktop and plain Linux Docker Engine.

2. Unified the DATABASE_STRING default into a single `DEFAULT_POSTGRES_DATABASE_STRING` constant in `llm_setup.py`, reused in the seed dict, local Postgres path, and Docker Postgres path. No more divergent defaults.

marcmuon (approved):
Both blocking items resolved in 0e977c3c — verified:

1. `extra_hosts: ["host.docker.internal:host-gateway"]` added to the `skyvern` service in `docker-compose.yml`. `host.docker.internal` will now resolve on plain Linux Docker Engine, so the `BROWSER_REMOTE_DEBUGGING_URL` written to `.env` works across Desktop and Engine.

2. `DOCKER_POSTGRES_DATABASE_STRING` renamed to `DEFAULT_POSTGRES_DATABASE_STRING`, moved to `llm_setup.py` so the seed dict can import it, and reused at all three call sites (`llm_setup.py:47`, `database.py:126`, `database.py:251`). No stale no-port literal remains in the CLI flow. Cleanly also addressed the non-blocking `database.py:128` literal.

LGTM. Approving.

PR #5734, opened 2026-04-30
fix(SKY-9283): auto-route browser type from profile source

LawyZheng:


LawyZheng:

Source PR: https://github.com/Skyvern-AI/skyvern-cloud/pull/10684

PR #5737, opened 2026-04-30
fix: treat 'already exists' as success in quickstart DB setup

suchintan:
## Summary
- When the pre-check for existing Postgres roles/databases fails to detect them (e.g. due to container timing), `createuser`/`createdb` would error with "already exists" and show a scary red warning during `skyvern quickstart`
- Now captures stderr directly via `subprocess.run` and treats "already exists" as a success condition, showing a green checkmark instead
- Genuine creation failures still surface the warning and fire analytics events

## Test plan
- [ ] Run `skyvern quickstart` with an existing `skyvern` role and database — should show green "already exists" messages
- [ ] Run `skyvern quickstart` on a fresh setup — should create user/database normally
- [ ] Verify a real failure (e.g. Docker not running) still shows the red warning


suchintan:
**Second commit: deferred heavy imports in `init_command.py`**

Investigated why `skyvern quickstart` shows no output after the "Detected Windows environment" config warning. The root cause is that `init_command.py` imported three heavy modules at the top level:

- `skyvern.forge.forge_app_initializer` → pulls in `anthropic`, `openai`, `fastapi`, the full LLM handler stack, artifact storage, DB layers (~39 transitive imports)
- `skyvern.utils.migrate_db` → pulls in `alembic`
- `.mcp` → pulls in `skyvern.forge.app`, security, DB schemas

All of these loaded eagerly before `quickstart()` could print its banner. Moved them to inline imports at their actual call sites so the CLI starts up fast and only loads heavy deps when the user reaches those wizard steps.

marcmuon (approved):

PR #5761, opened 2026-05-01
SKY-9422 Hide copilot workflow runs from history

andrewneilson:


docs/docs.json
wintonzheng, line 1: undo the change here.

PR #5777, opened 2026-05-01
Fix: include SDK artifacts dirs ignored by overly-broad gitignore

wintonzheng:
## Summary
The `artifacts/` pattern in `.gitignore` matched any `artifacts/` directory in the tree, including the Fern-generated SDK modules:
- `skyvern/client/artifacts/`
- `skyvern-ts/client/src/api/resources/artifacts/`

These get re-created on every Fern regen but were silently dropped by `git add` during version bumps. As of `1.0.32`, the auto-publisher run on `main` fails the TS build:

\`\`\`
src/Client.ts(4,27): error TS2307: Cannot find module './api/resources/artifacts/client/Client.js'
src/api/resources/index.ts(1,28): error TS2307: Cannot find module './artifacts/index.js'
\`\`\`

Failing run: https://github.com/Skyvern-AI/skyvern/actions/runs/25236452890/job/74003613624

## Changes
- `.gitignore`: anchor `artifacts/` to repo root (`/artifacts/`) so it only matches the top-level dir. Same fix as the prior `dd1e7ae4` commit, which was reverted by `e25c323d`.
- Check in the currently-missing artifact modules so the npm/PyPI publishers can find them on `main`.

## Test plan
- [ ] CI passes on this PR
- [ ] After merge, re-run the failing `Build Skyvern TS SDK and publish to npm` workflow and confirm it builds + publishes
- [ ] Verify Python SDK (`skyvern==1.0.32`) and TS SDK (`skyvern@1.0.32`) appear on PyPI / npm


marcmuon (approved):
Approve with suggestions — this fix is correct and matches how every other Fern-generated subpackage (`schedules/`, `scripts/`) is shipped. Verified:
- The seven Fern files match the OpenAPI spec on this branch (`GET /v1/artifacts/{artifact_id}/content`).
- Their parent files (`skyvern/client/__init__.py:52`, `skyvern/client/client.py:2480`, `skyvern-ts/client/src/Client.ts`, `skyvern-ts/client/src/api/resources/index.ts`) already reference `artifacts`, so this fills the missing leaf.
- The committed tree on main has zero other `artifacts/` directories under version control, so anchoring `/artifacts/` doesn't accidentally re-include anything.

Three non-blocking notes:

1. The PR description says `e25c323d` reverted `dd1e7ae4` — actually `e25c323d`'s .gitignore diff is `+log` only and it didn't touch any artifact files. PR #5453 was branched from main *before* dd1e7ae4 landed and silently clobbered both during merge. GitHub's compare view doesn't show this because the diff is taken against the merge parent, not the PR base. Worth flagging in the description so we can spot the next stale-base merge before it regresses this a third time.

2. After this lands, locally-generated `*/artifacts/` directories (`tests/artifacts/`, `scripts/artifacts/`, etc.) will start showing up in `git status`. Purely local noise — no committed dirs are affected. One-line heads-up in the description would prevent dev confusion.

3. While you're here, consider anchoring the rest of the `# Skyvern ignores` block (`/browser_sessions/`, `/videos/`, `/traces/`, `/har/`, `/temp/`, `/files/`). Same failure mode is one Fern resource-name collision away.

Out-of-scope context, NOT a blocker: the new Python `RawArtifactsClient.get_artifact_content` calls `_response.json()` on a route that returns binary bytes (PNG, octet-stream, video). It will always raise `ApiError` for non-text artifacts. This is an OpenAPI spec gap (route declares `application/json` for the 200 response) and predates this PR. Worth a follow-up ticket; please don't fold a fix into this release PR.

PR #5804, opened 2026-05-04
feat: add curl-based Skyvern server installer

marcmuon:
## Ticket

https://linear.app/skyvern/issue/SKY-9476/add-curl-installer-for-skyvern

## Description

Adds a one-line curl installer for the self-hosted Skyvern CLI/server path:

```sh
curl -LsSf https://install.skyvern.com | sh
```

The installer now targets the post-split package contract:

- `pip install skyvern` remains the lightweight Cloud/API SDK path and belongs in a project environment.
- `pip install "skyvern[local]"` remains the embedded local SDK path and belongs in a project environment.
- The curl installer is intentionally for the isolated CLI/server tool and installs `skyvern[server]` with `uv tool install --python '>=3.11,<3.14'`.

The script bootstraps `uv` if missing, installs the server extra as an isolated tool, and prints the next setup command. It no longer runs interactive setup by default because the canonical `curl | sh` invocation does not provide a real interactive stdin. Users can opt in with `--run-setup` / `--run-init`; when stdin is not a TTY the script skips setup and prints the command to run manually.

## Quickstart compatibility

This is compatible with the incoming quickstart install-type selector from Skyvern Cloud PR #10976. After installation, the script checks whether `skyvern quickstart --help` exposes `--install-type`:

- If available, it prints/runs `skyvern quickstart --install-type server`.
- Otherwise, it falls back to `skyvern quickstart`.

That keeps this PR independent of the exact OSS sync order while still using the narrower server quickstart once it lands.

## Hosting note

The script should be hosted from the OSS repo, not skyvern-cloud, because it installs the public OSS PyPI package. Suggested public endpoint: `install.skyvern.com` via Cloudflare Pages/Worker or Vercel pointed at this repo's `install.sh` on `main`. Raw GitHub can be a temporary fallback, but should not be the promoted URL.

## How Has This Been Tested?

- `sh -n install.sh scripts/test-install.sh`
- `git diff --check`
- `pre-commit run --files install.sh scripts/test-install.sh` -> shellcheck and generic hooks passed
- `sh install.sh --help`
- `sh install.sh --variant local --no-init` -> rejects with `pip install "skyvern[local]"` guidance
- `SKYVERN_RUN_INIT=wat sh install.sh --variant server --version 0` -> rejects invalid env override before install
- Mocked `PATH` smoke: verifies the installer calls `uv tool install --python >=3.11,<3.14 --force skyvern[server]`
- Mocked quickstart detection: verifies the success hint becomes `skyvern quickstart --install-type server` when that flag exists, and `skyvern quickstart` otherwise
- `uv pip install --python <temp venv python> --dry-run 'skyvern[server]'` -> resolved published `skyvern==1.0.35` server extra successfully

## Not tested

- Full Docker matrix in `scripts/test-install.sh`.
- Real `uv tool install 'skyvern[server]'`, to avoid replacing the local user-level `skyvern` tool during review.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -71,14 +71,13 @@ if ! command -v uv >/dev/null 2>&1; then
     # Download then execute (not `curl ... | sh`): POSIX /bin/sh has no
     # pipefail, so curl failures would otherwise be swallowed silently.
     UV_INSTALLER="$(mktemp -t skyvern-uv-installer.XXXXXX)"
+    trap 'rm -f "$UV_INSTALLER"' EXIT
     if ! curl -LsSf https://astral.sh/uv/install.sh -o "$UV_INSTALLER"; then
-        rm -f "$UV_INSTALLER"
         printf 'ERROR: failed to download uv installer from astral.sh.\n' >&2
         printf '       Check your network/proxy. See https://docs.astral.sh/uv/getting-started/installation/\n' >&2
         exit 1
     fi
     sh "$UV_INSTALLER"
-    rm -f "$UV_INSTALLER"
     export PATH="$HOME/.local/bin:$PATH"
     if ! command -v uv >/dev/null 2>&1; then
         printf 'ERROR: uv installer ran but uv is not on PATH.\n' >&2
--- a/install.sh
+++ b/install.sh
@@ -70,7 +70,11 @@ if [ -n "$SKYVERN_VERSION" ]; then
 fi
 
 printf 'installing %s via uv tool install...\n' "$PKG_SPEC"
-uv tool install --force "$PKG_SPEC"
+# Pin Python to a version inside skyvern's `requires-python = ">=3.11,<3.14"`.
+# Without this, uv may pick the host's Python 3.14 (or newer), in which case
+# environment-marker dependencies like playwright (`<'3.14'`) silently drop
+# out of the resolution and the CLI crashes at import time.
+uv tool install --python 3.13 --force "$PKG_SPEC"
 
 if [ "$SKYVERN_RUN_INIT" = "1" ]; then
     printf "\nrunning 'skyvern init' (installs Chromium, ~150MB)...\n"
--- a/install.sh
+++ b/install.sh
@@ -1,28 +1,8 @@
 #!/bin/sh
-# Skyvern installer.
-#
-# Usage:
-#   curl -LsSf https://install.skyvern.com | sh
-#   curl -LsSf https://install.skyvern.com | sh -s -- --no-init
-#   curl -LsSf https://install.skyvern.com | sh -s -- --version 1.0.31
-#
-# Environment overrides:
-#   SKYVERN_RUN_INIT=0   skip `skyvern init` (Chromium browser install)
-#   SKYVERN_VERSION=X    pin a specific PyPI version
-#
-# What this does:
-#   1. Bootstraps `uv` (Astral's Python toolchain) if not already installed.
-#      uv handles Python version management — the host machine doesn't need
-#      a pre-installed Python.
-#   2. Installs the `skyvern` package from PyPI as an isolated tool, exposing
-#      the `skyvern` command on PATH.
-#   3. Runs `skyvern init` to install Chromium for browser automation
-#      (skip with --no-init when running in CI / minimal containers).
+# Skyvern installer. Run `install.sh --help` for usage.
 
 set -eu
 
-# Empty SKYVERN_RUN_INIT means "use the variant's default"; --no-init or
-# `SKYVERN_RUN_INIT=0` overrides to 0; `SKYVERN_RUN_INIT=1` forces it on.
 SKYVERN_RUN_INIT="${SKYVERN_RUN_INIT:-}"
 SKYVERN_VERSION="${SKYVERN_VERSION:-}"
 SKYVERN_VARIANT="${SKYVERN_VARIANT:-server}"
@@ -35,8 +15,8 @@ while [ $# -gt 0 ]; do
         --variant)   SKYVERN_VARIANT="${2:?--variant requires a value}"; shift 2 ;;
         --variant=*) SKYVERN_VARIANT="${1#--variant=}"; shift ;;
         --help|-h)
-            # Self-parsing of "$0" doesn't work when piped (`curl ... | sh`);
-            # `$0` is then the shell name, not a real file. Embed help inline.
+            # Heredoc rather than self-parsing $0: when piped, $0 is the
+            # shell name, not a real file path.
             cat <<'EOF'
 Skyvern installer.
 
@@ -59,13 +39,12 @@ Environment overrides:
   SKYVERN_VERSION=X    same as --version
 
 What this does:
-  1. Bootstraps 'uv' (Astral's Python toolchain) if not already installed.
-     uv handles Python version management — the host machine doesn't need
-     a pre-installed Python.
-  2. Installs the matching PyPI package as an isolated tool, exposing
-     the 'skyvern' command on PATH.
-  3. For variants that need a browser, runs 'skyvern init' to download
-     Chromium (skip with --no-init when running in CI / minimal containers).
+  1. Bootstraps 'uv' if missing — uv manages Python so the host
+     doesn't need one preinstalled.
+  2. Installs the variant's PyPI package as an isolated tool
+     ('skyvern' on PATH).
+  3. For variants that need a browser, runs 'skyvern init' to fetch
+     Chromium (~150MB; skip with --no-init).
 EOF
             exit 0
             ;;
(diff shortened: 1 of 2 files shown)


marcmuon:
**Single-machine smoke test — passed (after fix in 43e3f435).**

**Host:** macOS 14, arm64. uv 0.11.7 already installed.

**Run 1 (initial commit c54a1932):** `bash install.sh --no-init` succeeded but `skyvern --version` failed with `ModuleNotFoundError: No module named 'playwright'`. Root cause: uv selected Python 3.14.4 for the tool venv (the host had it available), which falls outside skyvern's `requires-python = ">=3.11,<3.14"`. Both playwright deps have markers `<'3.14'`, so they silently dropped from the resolution. uv 0.11.7 should have rejected the install on `requires-python` alone but didn't.

**Fix (43e3f435):** added `--python 3.13` to `uv tool install` so the tool venv lands inside the supported range regardless of which Python versions are on the host.

**Run 2 (post-fix):** `bash install.sh --no-init` completes in ~3s (cached). Tool venv is Python 3.13.9. `skyvern --version` executes (Rich table renders fully). `uv tool uninstall skyvern` cleans up; `command -v skyvern` confirms removal.

**Not yet tested:**
- Full Docker matrix (`scripts/test-install.sh`) — needs CI runner with Docker.
- `--no-init=0` path with real Chromium download.
- Fresh-machine bootstrap path (uv not pre-installed).

marcmuon:
**Code review applied + earlier smoke-test claim corrected.**

Pushed `fdc02218`:

1. **[Major from review]** `install.sh` line 56 (was): `curl ... | sh` is now `curl -o tmpfile && sh tmpfile` with explicit failure detection on the curl step and a `command -v uv` sanity check after. POSIX `/bin/sh` has no `pipefail`, so the previous form swallowed curl failures.
2. **[Minor from review]** `install.sh` `--help` was self-parsing `$0`, which is broken under the canonical `curl ... | sh -s -- --help` invocation (`$0` is then `"sh"`, not a real path). Replaced with an inline heredoc — verified working both directly and via pipe.
3. **[Minor from review]** `install.sh` `--python 3.13` swapped to `--python '>=3.11,<3.14'`. Verified uv 0.11.7 accepts the range syntax and resolves to the highest in-range Python on the host. This now tracks skyvern's `requires-python` automatically as it evolves.
4. **Correction to my prior comment:** I claimed `skyvern --version` "rendered the Rich table cleanly" — that was actually typer's *error* table for "No such option: --version". skyvern's CLI doesn't expose a `--version` flag at all. Updated `scripts/test-install.sh` to assert `skyvern --help >/dev/null` (exit 0) instead, which is the correct post-install signal that the package imports cleanly (playwright included).

Smoke test re-run on host (Mac arm64) post-fix:
- `bash install.sh --no-init` → 2s install, tool venv on Python 3.12.13 (highest in range)
- `skyvern --help` → exit 0
- `uv tool uninstall skyvern` → clean
- Direct `--help` and piped `cat install.sh | sh -s -- --help` both produce identical output.

Remaining items from review filed for follow-up (not in this PR):
- Pin uv-installer URL to a tagged release rather than `astral.sh/uv/install.sh` latest, for supply-chain hygiene.
- `scripts/test-install.sh` add a "uv preinstalled" image variant.
- SELinux/podman bind-mount label (`:ro,Z`) for non-Mac Linux CI.
- Filing an issue against skyvern's `pyproject.toml` re: playwright env-marker / `requires-python` mismatch.

marcmuon:
**Tri-model review applied — variant policy refactor (079a90c4).**

Reviewed against the March 2026 SDK package split PRD using Claude + Codex (gpt-5.5) + Gemini in parallel via `omc ask`. Codex read both files and cited line numbers; Gemini's review was opinion-only (workspace sandbox blocked the file paths) but useful on UX framing.

**Where they agreed:**
- Default `--variant` should be `server` (curl-installer audience are server/MCP/tire-kickers, not SDK consumers).
- `skyvern init` (Chromium) needs to be variant-aware — wrong for the post-split lightweight SDK.
- Single canonical script with variant policy is the right shape; subpath URLs are an optional thin-alias layer for marketing/telemetry.
- Pre-split success message (`Try: skyvern --help`) is too generic for a multi-variant world.

**Codex-specific catches (file-grounded):**
- Tests had zero variant dimension (`scripts/test-install.sh:27, :38`). Critical pre-split gap.
- The earlier comment claiming the Python range "tracks `requires-python` automatically" was misleading — the range is hardcoded.
- Don't infer Python range from PyPI metadata at runtime; extras can't differ in `requires-python`, so the JSON metadata becomes ambiguous post-split. Hardcode per-variant + add CI diff check.
- Don't ship the public `--variant {sdk,mcp,...}` surface before those PyPI targets exist — but DO refactor internals now.

**Gemini-specific framing:**
- Intent-based variant naming (`self-host` vs `server`) reads better in prose; recommended keeping that for docs/README pathfinder tables, not for flag values.
- Subpath URLs (`install.skyvern.com/server`) are more "pasteable" in READMEs than `--variant` flags.

**Tier 1 landed in this PR (079a90c4):**
- `install.sh`: per-variant `case` table encoding `PKG_SPEC`, `PYTHON_SPEC`, `DEFAULT_RUN_INIT`, `SUCCESS_HINT`. Today only `--variant server` is accepted; reserved values (`sdk`, `mcp`, `cli`, `cloud`, `all`) reject with a clear "not yet available" message so the public flag API stays stable through the split.
- `install.sh`: `skyvern init` is driven by `DEFAULT_RUN_INIT` from the variant's policy, not a global default. `--no-init` / `SKYVERN_RUN_INIT` still override.
- `install.sh`: success message uses `$SUCCESS_HINT`.
- `install.sh`: misleading "tracks requires-python automatically" comment removed; new comment explicitly says the range is hardcoded and must be bumped intentionally.
- `scripts/test-install.sh`: `VARIANTS` dimension added (today just `server`). Split day is a data-only change.

**Verification:** POSIX syntax clean, `--help` documents `--variant`, reserved/unknown variant rejection works (exit 2), bare-default install path is byte-identical in behavior to before (2s install, `skyvern --help` exit 0, clean uninstall via `uv tool uninstall`).

**Tier 2 — explicit follow-ups (not in this PR):**
- Expand variant `case` arms when `skyvern[server]` extras and `skyvern-mcp` PyPI packages exist.
- Subpath aliases at `install.skyvern.com/{server,mcp,sdk}` (thin scripts that set `SKYVERN_VARIANT` and exec the canonical script).
- CI job that diffs each variant's `PYTHON_SPEC` against the published `requires-python` for that PyPI target.
- README "Pathfinder" table mapping intent → curl one-liner.
- Migration detection: re-running the installer against a pre-split monolithic install should print an upgrade note explaining the split.

install.sh
```diff
@@ -0,0 +1,140 @@
...
+        ;;
+esac
+
+uv_was_freshly_installed=0
+if ! command -v uv >/dev/null 2>&1; then
+    uv_was_freshly_installed=1
+    printf 'installing uv (one-time, into ~/.local/bin)...\n'
+    # Download then execute (not `curl ... | sh`): POSIX /bin/sh has no
+    # pipefail, so curl failures would otherwise be swallowed silently.
+    UV_INSTALLER="$(mktemp -t skyvern-uv-installer.XXXXXX)"
```
AronPerez, line 77: with `set -eu` a non-zero exit from L80 skips the `rm` on L81 and leaks the tmp file - `trap 'rm -f "$UV_INSTALLER"' EXIT` right here would make cleanup unconditional, wdyt?

scripts/test-install.sh
```diff
@@ -0,0 +1,58 @@
...
+failed=""
+for variant in $VARIANTS; do
+    for img in $IMAGES; do
+        printf '\n=== %s / variant=%s ===\n' "$img" "$variant"
+        if docker run --rm -v "$INSTALL_SH:/tmp/install.sh:ro" "$img" sh -c "
+            set -e
+            export DEBIAN_FRONTEND=noninteractive
+            apt-get update -qq >/dev/null
+            apt-get install -qq -y curl ca-certificates >/dev/null
+            sh /tmp/install.sh --variant '$variant' $init_flag
```
AronPerez: a `$variant` containing `'` would break out of the `'$variant'` quoting here - safer to pass through `-e VARIANT="$variant"` and read `"$VARIANT"` inside a single-quoted body, wdyt?

marcmuon: nice yeah changed https://github.com/Skyvern-AI/skyvern/pull/5804/changes/ca4bbc0d88c64bc1cdc4ec9856a981bc00b19cfb#diff-84e4f528f76a2e8f035630b952a9ec1b9ef4b017d7e472fc2ccc7eb132bfa18cR43

install.sh
marcmuon: added trap https://github.com/Skyvern-AI/skyvern/pull/5804/changes/ca4bbc0d88c64bc1cdc4ec9856a981bc00b19cfb#diff-043df5bdbf6639d7a77e1d44c5226fd7371e5259a1e4df3a0dd5d64c30dca44fR74

AronPerez (approved):

PR #5833, opened 2026-05-06
SKY-9488: make base SDK import safe

marcmuon:


marcmuon:
superseded by 5837 merge

PR #5836, opened 2026-05-06
Deprecate Streamlit credential path

suchintan:
## Summary
- replace generated .streamlit credentials with .skyvern/credentials.toml
- update Docker Compose, Kubernetes manifests, and self-hosted docs for the new credential path
- let explicit VITE_SKYVERN_API_KEY values take precedence over generated local credentials in the UI entrypoint
- keep skyvern doctor migration support for legacy .streamlit/secrets.toml installs

## Tests
- bash -n entrypoint-skyvern.sh entrypoint-skyvernui.sh
- python3 -m py_compile skyvern/cli/doctor.py skyvern/config.py
- docker compose config --quiet
- git diff --check
- pre-commit run --files skyvern/cli/doctor.py tests/unit/test_cli_doctor.py kubernetes-deployment/backend/backend-deployment.yaml
- uv run --extra server pytest -q tests/unit/test_cli_doctor.py


marcmuon (approved):
LGTM with a few non-blocking notes for follow-up.

**Wins**
- Heartbeat healthcheck is a real readiness signal — fixes a latent race where the old `test -f .streamlit/secrets.toml` probe reported healthy before `python -m skyvern.forge` was actually serving (file is written by the entrypoint before the API starts).
- All `.streamlit` references on the branch are intentional (deprecation/migration only).
- `.dockerignore`/`.gitignore` patterns don't collide with the `skyvern/` Python package.

**Follow-ups (non-blocking)**

🟡 **k8s readinessProbe wasn't migrated.** `kubernetes-deployment/backend/backend-deployment.yaml:36-40` still uses `test -f /app/.skyvern/credentials.toml` — the same file-existence pattern this PR replaced for compose. The pod can be marked Ready while the API isn't serving. Worth swapping to an `httpGet` against `/api/v1/heartbeat`.

🟡 **`_fix_legacy_streamlit_secrets` unconditionally unlinks.** `skyvern/cli/doctor.py:807` deletes the legacy file even on the path where `not backend_key and not legacy_key` (file existed but had no `cred=`). The check warns "no cred value found" but the fix still deletes — surprising asymmetry. Consider skipping the unlink when `legacy_key` is empty so the user can inspect it.

🟢 **Behavior change in `entrypoint-skyvernui.sh` priority** (env var now wins over credentials file, was the reverse) is correct for the new flow but worth a line in the PR body for anyone relying on the old precedence.

🟢 **`_check_api_key_consistency` no longer flags backend↔generated divergence.** With `canonical = backend_key or generated_key or legacy_key`, a stale generated file is silently shadowed when backend `.env` is set. Probably fine (backend is the new source of truth) but the docstring could note that generated is a fallback only.

🟢 **Redundant `import shutil` at `doctor.py:761`** — already imported at module top and used without re-import at line 802.

🟢 **No pytest coverage for the new doctor logic.** Given the file-deletion side effect, a small test over the four detection branches (no file / no cred / mismatch / match) would be cheap insurance.

🟢 **`_check_legacy_streamlit_secrets` always emits a row** even on clean installs. Minor noise; could early-return when the file doesn't exist.

suchintan:
Addressed the review notes in the latest push:

- changed the Kubernetes backend readiness probe to hit `/api/v1/heartbeat` instead of checking for the generated credentials file
- kept unparseable legacy `.streamlit/secrets.toml` files in place for inspection during `skyvern doctor --fix`
- tightened legacy `cred` parsing so keys like `not_cred` are not treated as credentials
- documented that generated credentials are only a fallback in the doctor consistency check
- removed the redundant `shutil` import
- added focused unit coverage for the legacy Streamlit doctor branches
- updated the PR body to call out the UI env-var precedence change

PR #5851, opened 2026-05-06
fix(SKY-9621): add pool_timeout/pool_recycle config and raise prod API pool size

AronPerez:


AronPerez:
Addressed @copilot-pull-request-reviewer's comment on `css-vars.test.ts:20`:

- Replaced `__dirname` (undefined in ESM since `skyvern-frontend` is `"type": "module"`) with an ESM-safe derivation via `fileURLToPath(import.meta.url)`.
- Also filtered `CSS_FILES` through `existsSync` so the suite runs in the OSS sync target where `cloud/` and `eval/` entrypoints don't exist (this was the actual CI failure).

Verified locally — vitest passes. Pushed in 41fb4449.

PR #5908, opened 2026-05-08
[CLI] Improve quickstart secret prompt feedback

suchintan:
## Summary
- Add a masked secret prompt helper that echoes `*` characters while preserving secret input values.
- Use the helper for quickstart/init LLM and Skyvern API key prompts.
- Cover paste masking, backspace behavior, and stream handling with focused unit tests.

## Root Cause
Rich/Typer password prompts delegate to `getpass`, which intentionally echoes nothing. That makes pasted API keys look like they were not accepted during quickstart.

## Validation
- `ruff check skyvern/cli/masked_prompt.py skyvern/cli/llm_setup.py skyvern/cli/init_command.py tests/unit/test_masked_prompt.py`
- `ruff format --check skyvern/cli/masked_prompt.py skyvern/cli/llm_setup.py skyvern/cli/init_command.py tests/unit/test_masked_prompt.py`
- `uv run --extra server --group dev pytest tests/unit/test_masked_prompt.py`
- PTY smoke test for masked terminal input

### DIFF
--- a/skyvern/cli/init_command.py
+++ b/skyvern/cli/init_command.py
@@ -16,6 +16,7 @@ from .browser import setup_browser_config
 from .console import console
 from .database import setup_postgresql
 from .llm_setup import setup_llm_providers, update_or_add_env_var
+from .masked_prompt import ask_secret
 
 
 def init_env(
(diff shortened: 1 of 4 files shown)


marcmuon:
Request changes: I think there is one real correctness blocker in the new masked prompt reader.

`skyvern/cli/masked_prompt.py` drops only the `ESC` byte for terminal escape sequences, then captures the printable remainder. Repro against this PR head:

```text
'\x1b[200~sk-abc\x1b[201~\n' => '[200~sk-abc[201~'
'ab\x1b[D\n' => 'ab[D'
```

That means bracketed paste in modern terminals can corrupt the API key captured by the new `ask_secret` quickstart/init paths while the UI still renders stars and appears successful.

Suggested fix: consume full ANSI/CSI sequences before appending to the secret buffer, especially bracketed paste start/end (`ESC [ 200 ~`, `ESC [ 201 ~`), and add a regression test for pasted bracketed input.

I would not block on the `mcp_commands.py` password prompts in this PR. They are outside this diff and under the separate `skyvern mcp` surface; that path also sanitizes ANSI sequences via `_sanitize_prompt_response`, so it is a consistency follow-up rather than the same corruption bug.

marcmuon (approved):

PR #5954, opened 2026-05-12
fix(scraper): recover from multi-hop navigation context destruction (SKY-9470)

AronPerez:


pedrohsdb (approved):
lgtm — cloud parent skyvern-cloud#11011 was approved and merged

PR #5957, opened 2026-05-12
fix(workflow): JSON-encode dict/list in trigger payload (SKY-9766)

celalzamanoglu:


celalzamanoglu:
The SKY-9766 changes (jinja JSON finalize for workflow trigger payloads) already landed on main via other sync PRs. This PR got stuck due to a CI issue and after rebasing only had an unrelated diff left. Reset the branch to main to auto-close.

PR #5959, opened 2026-05-12
Pre-build UI assets in Docker image instead of at container startup (SKY-7342)

celalzamanoglu:


pedrohsdb (approved):
lgtm — cloud parent skyvern-cloud#9115 was approved and merged

PR #6010, opened 2026-05-15
Update 2FA docs to reference workflows

suchintan:


andrewneilson (approved):
Reviewed: docs updates correctly reflect the actual webhook contract in `skyvern/services/otp_service.py` (`workflow_run_id` in request body, response only reads `verification_code`). Examples switched from `client.login()` to `client.run_workflow()` are aligned with the SDK signature. The `Mintlify Deployment` addition to `ignore-checks` in auto-merge-sync.yml is consistent with that check landing as SKIPPED for non-docs branches.

PR #6025, opened 2026-05-17
[CLI] Add non-interactive quickstart setup

suchintan:
## Description

Adds an agent-friendly/non-interactive path for self-hosted quickstart while preserving the latest quickstart install-type flow.

- Adds `--non-interactive`, `--yes`, prompt-skipping, browser, analytics, and service-start flags to `skyvern quickstart`
- Adds non-interactive hooks to `init_env` for local server setup
- Makes missing server dependency installs more resilient with pip retries/timeouts
- Seeds local org/API key without constructing the full Forge runtime during setup
- Adds regression coverage for unattended quickstart options

cc @marcmuon

## How Has This Been Tested?

- `uv run ruff check skyvern/cli/browser.py skyvern/cli/init_command.py skyvern/cli/mcp.py skyvern/cli/quickstart.py skyvern/exceptions.py tests/unit/test_init_mcp.py tests/unit/test_quickstart_server_extra.py`
- `uv run ruff format --check skyvern/cli/browser.py skyvern/cli/init_command.py skyvern/cli/mcp.py skyvern/cli/quickstart.py skyvern/exceptions.py tests/unit/test_init_mcp.py tests/unit/test_quickstart_server_extra.py`
- `uv run --extra server pytest tests/unit/test_quickstart_server_extra.py tests/unit/test_init_mcp.py tests/unit/test_exception_messages.py::test_require_server_extra_modules_requires_server_sentinels tests/unit/test_exception_messages.py::test_require_server_extra_modules_catches_partial_server_graph tests/unit/test_exception_messages.py::test_require_server_extra_modules_discriminates_from_local_with_explicit_modules`
- `uv run python -m skyvern.cli.commands quickstart --help`
- `uv run python -m skyvern.cli.commands init --help`

N/A


skyvern/cli/quickstart.py
```diff
@@ -668,12 +689,66 @@ def quickstart(
...
     env_scope_value = env_scope if isinstance(env_scope, str) else None
+    if non_interactive:
+        install_type_value = install_type_value or QuickstartPath.SERVER.value
+        yes = True
+        skip_llm_setup = True
+        skip_mcp = True
+        browser_type = browser_type or "chromium-headless"
+        analytics_id = analytics_id or "anonymous"
+        if start_services_now is None:
+            start_services_now = False
```
marcmuon, line 751: nit/follow-up: this block forces every other prompt-skipping flag, but not `no_postgres`. If the user runs `skyvern quickstart --non-interactive` on a Docker host without a `postgresql-container` and without `--database-string`, `setup_postgresql` still hits `Confirm.ask` at `skyvern/cli/database.py:208` and hangs on a TTY-less agent runner.

Two equally fine fixes:
- auto-set `no_postgres = True` here when `database_string` is empty (and emit a guidance panel pointing at `--database-string`, like the docker-compose branch a few lines below does), or
- tighten the help text on `--non-interactive` to say `--database-string` is required for the server path.

The documented path (`--non-interactive --database-string postgresql+psycopg://...`) already works — this is just to keep the bare `--non-interactive` invocation from silently hanging.

skyvern/cli/init_command.py
```diff
@@ -385,27 +422,16 @@ def set_env_var(key: str, value: str) -> None:
...
-        def _load_setup_local_organization() -> Callable[[], Coroutine[Any, Any, str]]:
-            from .mcp import setup_local_organization  # noqa: PLC0415
-
-            return setup_local_organization
-
-        start_forge_app = _run_with_server_dependency_install(_load_start_forge_app)
-        setup_local_organization = _run_with_server_dependency_install(_load_setup_local_organization)
-        _run_with_server_dependency_install(start_forge_app)
-        api_key = _run_with_server_dependency_install(lambda: asyncio.run(setup_local_organization()))
+        api_key = asyncio.run(_setup_local_organization_from_database())
```
marcmuon: nit/follow-up: small UX regression vs the pre-PR shape — `migrate_db` on the line above goes through `_run_with_server_dependency_install`, but this `asyncio.run(...)` doesn't. If a server-extra sub-dependency (sqlalchemy, asyncpg, psycopg, etc.) fails to import mid-flight, the user gets a raw stack trace here instead of the friendly install panel.

Suggested:
```python
api_key = _run_with_server_dependency_install(
    lambda: asyncio.run(_setup_local_organization_from_database())
)
```

marcmuon (approved):
LGTM. Agent-friendly path is a clear improvement and the new typed `init_env` overloads keep the existing callers working.

Two small follow-ups I'd suggest landing in a follow-up (or here if easy) — left inline:

1. `--non-interactive` doesn't propagate to `no_postgres`, so on a Docker host with no `postgresql-container` and no `--database-string` the flow still blocks at `setup_postgresql`'s `Confirm.ask`. The docker-compose branch already tells users to pass `--database-string` for unattended runs, so either auto-set `no_postgres=True` here or tighten the help text to make `--database-string` required under `--non-interactive`.
2. The new direct `asyncio.run(_setup_local_organization_from_database())` skips the `_run_with_server_dependency_install` wrapper that the rest of the flow uses, so if a server-extra import fails partway through, the user gets a raw stack trace instead of the friendly install panel.

Not blocking either way — happy to ship.

PR #6159, opened 2026-05-25
Add packaged UI pip install flow

suchintan:
## Summary

This PR adds a Node-free packaged UI path for pip installs by introducing a same-version `skyvern-ui` package containing prebuilt frontend assets. Users can now choose a lightweight SDK install, backend-only install, UI-only install, or local server + packaged UI install via extras.

## What changed

- Add `skyvern[ui]` and `skyvern[all]` extras, with a same-version `skyvern-ui` package for prebuilt frontend assets.
- Add Python runtime support for serving packaged UI assets when a source frontend checkout is unavailable.
- Add interactive and non-interactive UI auto-install paths for `skyvern run ui`, `skyvern run all --install-ui`, and quickstart.
- Make pip quickstart SQLite-first by default, with `--postgres` preserving the local Postgres-container path.
- Harden the packaged artifact server with loopback binding, tokenized artifact URLs, restricted CORS, and root-limited file access.
- Update release/CI workflows to build and publish both packages and run pip package smoke tests on Python 3.11 and 3.13.
- Update README/docs for the new install paths and add unit coverage for packaging, quickstart, and UI runtime behavior.

## Install paths

- `pip install skyvern`: lightweight Cloud/API SDK usage.
- `pip install "skyvern[ui]"`: packaged UI assets only.
- `pip install "skyvern[server]"`: backend-only local server usage.
- `pip install "skyvern[all]"`: local server + packaged UI.

## Validation

- `uv run pytest tests/unit/test_init_mcp.py tests/unit/test_quickstart.py tests/unit/test_quickstart_command.py tests/unit/test_quickstart_browser_install.py tests/unit/test_run_commands_ui.py tests/unit/test_ui_runtime.py tests/unit/test_pyproject_extras.py -q`
- `uv run ruff check skyvern/cli/init_command.py skyvern/cli/quickstart.py skyvern/cli/run_commands.py skyvern/cli/ui_runtime.py tests/unit/test_init_mcp.py tests/unit/test_quickstart.py tests/unit/test_quickstart_command.py tests/unit/test_run_commands_ui.py tests/unit/test_ui_runtime.py tests/unit/test_pyproject_extras.py`
- `uv run mypy skyvern/cli/init_command.py skyvern/cli/quickstart.py skyvern/cli/run_commands.py skyvern/cli/ui_runtime.py`
- `uv lock --check`
- `git diff --check`

Docker note: the Python 3.13 Docker smoke rerun was attempted after the final artifact-runtime changes, but Docker daemon was unavailable locally. The CI matrix now covers this path for Python 3.11 and 3.13.


wintonzheng (approved):
nice work — approving. a few things worth a follow-up:

**`skyvern[ui]` as a standalone extra** — what's the use case? without `[server]` the bundle is rendered with `VITE_API_BASE_URL=http://localhost:8000` and an empty `SKYVERN_API_KEY`, and there's no `--api-url` flag on `skyvern run ui` to point it elsewhere. happy to be wrong, but right now i can't think of a scenario where someone wants only `[ui]`. consider folding it into `[all]` and dropping the standalone extra, or wiring up remote-API config as a follow-up.

**security in `ui_runtime.py`:**

1. *Token leak via `Referer`* — the artifact token is in the URL path (`/{token}/artifact/...`). any served artifact (JSON/HTML/SVG) that triggers an external subresource will leak the full URL via `Referer` to a third-party host. fix: set `Referrer-Policy: no-referrer` on every artifact response, or move the token to a header.

2. *`_cors_origin` echoes `localhost:<port>` when `Origin` is missing* (`skyvern/cli/ui_runtime.py`, the `_cors_origin` helper). token still gates access today so not exploitable, but it's a footgun if anyone later adds a write endpoint. return `None` when `Origin` is absent.

3. *Cache-dir permissions* — `prepare_installed_ui_dist` writes the `SKYVERN_API_KEY` into JS/HTML at `~/.cache/skyvern/ui/<version>/runtime/` with default umask. fine on a single-user laptop, but on a shared box another local user can `cat` it. `mkdir(mode=0o700)` + `chmod 0o600` on rendered files is cheap. (the same key already lives in `.env`, so this is "nice to fix", not blocking.)

**release ordering** (`.github/workflows/sdk-release.yml`) — if `skyvern` uploads and then `skyvern-ui` upload fails (transient PyPI 5xx), `pip install "skyvern[ui]==<new>"` is broken for whoever resolves between the two uploads. publish `skyvern-ui` first, then `skyvern`, so the pinned dep is resolvable before the parent ships.

**`run all` silent degrade in non-interactive mode** (`run_commands.py` → `_install_packaged_ui_if_requested`) — when not a TTY and `skyvern-ui` isn't installed, it returns False and `run all` starts backend-only with no warning. log a structured warning at minimum so docker/systemd users notice.

**nit:** `_install_packaged_ui_if_requested` shells `[sys.executable, "-m", "pip", "install", ...]`. same pattern was already in `quickstart.py`, but with a second invocation site it's more reachable — in a uv-managed venv this either fails (no pip) or desyncs `uv.lock`. detect uv and prefer `uv pip install`, or print "run this yourself" with the command. not a blocker for this PR since it was already there.

**`file://` smuggling** (`_normalize_artifact_path`) — `_normalize_artifact_path("file:///etc/passwd")` returns `/etc/passwd`. containment still rejects it as long as no configured root contains `/etc`, but `ARTIFACT_STORAGE_PATH=/tmp` is a common setting → reachable for `/tmp/<anything>`. reject `file://` outright.

**tests** — none of the security claims (Referer, CORS-missing-Origin, cache perms, `file://`) have negative-path coverage today. worth adding.

PR #6170, opened 2026-05-26
fix(SKY-10194): show credential TOTP identifier in placeholder instead of pre-filling

trevor-cheung:


trevor-cheung:

PR #6172, opened 2026-05-26
fix(deps): pin typer<0.26 and declare click explicitly

trevor-cheung:


trevor-cheung:

PR #6191, opened 2026-05-27
docs(proposal): Firefox-based stealth backend

feder-cr:
opening as draft to check interest before building the integration.

would a Firefox-based stealth backend option be in scope? parallel to the stealth-chromium option being proposed in PR #5340 (CloakBrowser), added as `BROWSER_TYPE=stealth-firefox` with the same optional-dependency shape.

backend exists and works today: feder-cr/invisible_firefox (MPL-2, same license as Firefox upstream, patches at the C++ source level so there are no JS shims to detect). Firefox as a second engine is useful for sites where Chromium UA gets flagged by default.

this PR only adds a stub docs page so the proposal has somewhere concrete to land. tracking discussion: #6190

if the answer is "not in scope" i'll close it without noise.

### DIFF
--- a/docs/proposals/invisible-firefox-backend.mdx
+++ b/docs/proposals/invisible-firefox-backend.mdx
@@ -1,0 +1,24 @@
+---
+title: "Firefox-based stealth backend (proposal)"
+description: "Draft proposal for a stealth-firefox BROWSER_TYPE option using invisible_playwright"
+---
+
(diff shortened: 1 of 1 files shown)


feder-cr:
@wintonzheng is a Firefox-stealth backend something Skyvern would want for targets that block the default browser? draft's just a proposal doc for now, easy to close if not in scope.

PR #6202, opened 2026-05-27
fix(SKY-10325): reopen credential modal after Esc dismiss

trevor-cheung:


trevor-cheung:
Superseded by the fixes made in #6205.

PR #6209, opened 2026-05-27
Update API specifications with fern api update

wintonzheng:
Update API specifications by running fern api update.

### DIFF
--- a/fern/openapi/skyvern_openapi.json
+++ b/fern/openapi/skyvern_openapi.json
@@ -55,7 +55,7 @@
           "content": {
             "application/json": {
               "schema": {
-                "$ref": "#/components/schemas/TaskRunRequest"
+                "$ref": "#/components/schemas/TaskRunRequest-Input"
               }
             }
           }
@@ -177,7 +177,7 @@
           "content": {
             "application/json": {
               "schema": {
-                "$ref": "#/components/schemas/WorkflowRunRequest"
+                "$ref": "#/components/schemas/WorkflowRunRequest-Input"
               }
             }
           }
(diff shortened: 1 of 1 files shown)


trevor-cheung:
Reviewed the regen — auto-generated by `fern api update`, single file (+1503 / −345). Surface changes captured below.

## New / changed endpoints

| Endpoint | Source |
|---|---|
| `POST /v1/runs/cancel` (`bulk_cancel_runs`) | new — body `{ run_ids: string[] }`, `maxItems: 100`, response splits `cancelled` / `failed` |
| `POST /v1/workflows/runs/{workflow_run_id}/retry` | from #6195 (regen of existing route, now tagged under both `Workflow Runs` + `Workflows`) |
| `GET /v1/version` | new — returns `Record<string,string>` |
| `PATCH /v1/browser_profiles/{profile_id}` (`update_browser_profile`) | new — optional `name` / `description` |
| `206 Partial Content` / `416 Range Not Satisfiable` on artifact download | new responses, indicates HTTP Range support added server-side |

## Schema changes

- `TaskRunRequest` / `WorkflowRunRequest` split into `…-Input` / `…-Output` variants (Pydantic v2 input-vs-output divergence). Refs updated at all call sites.
- `max_elapsed_time_minutes` added across run-request schemas (from #6201).
- `USER_DEFINED_ERROR` added as a const variant in the error-type union.
- Extended `proxy_location` description with custom proxy URL dict format for self-hosted deployments.

## Risks / non-blocking follow-ups

1. **`max_elapsed_time_minutes` constraints are inconsistent across routes.** Some sites declare `integer | null` with no bounds; at least one declares `minimum: 1, maximum: 240`. Suggests the Pydantic models on different routes don't share a single constrained type. The 1–240 cap (4-hour ceiling) should probably apply everywhere — worth aligning the source models so the next regen converges.

2. **`GET /v1/version` response is untyped** — `additionalProperties: string` means SDK consumers get `Dict[str, str]` with no documented key. Cheap win to define a typed `VersionResponse` with a named field (e.g. `version` or `sha`).

3. **`TaskRunRequest` → `TaskRunRequest-Input` rename is a potential break for external OpenAPI consumers** (not fern-generated SDKs — those handle it). Worth checking whether any pinned Postman collections / docs / customer-side codegen reference the old names.

4. **`BulkCancelRunsRequest.run_ids` has `maxItems: 100` but no `minItems`** — empty array should presumably short-circuit; worth confirming the handler doesn't open a transaction for zero work.

5. **No `caller_type` exposure** (from #6205) — confirm that's intentional (column is internal-only) vs the regen running before the route surfacing landed.

Approve overall — the spec faithfully reflects the recent backend PRs. The items above are upstream alignment work for the next regen, not blockers for this one.

PR #6350, opened 2026-06-03
fix: run Code Blocks in the agent's authenticated browser context

SGudbrandsson:
## Description
A workflow Code Block (running JS via the Playwright `page` — `page.goto` /
`page.evaluate`) does not run in the same authenticated browser context as the
preceding agent blocks. After a `login`/`navigation` block authenticates, the
next Code Block's `page.goto(<authed url>)` lands on the sign-in page, because
the block is handed a separate, unauthenticated browser. Reproduces with a
`login → code` (or `login → navigation → code`) workflow, with
`persist_browser_session` both on and off.
**Root cause.** Introduced with the v1.0.34 "script execution with browser
session state persistence" work. In a workflow run every browser-acquisition
path registers the authenticated `BrowserState` under the `workflow_run_id`
(agent tasks via `RealBrowserManager.get_or_create_for_task`, script bootstrap
via `get_or_create_for_workflow_run`). But Code Blocks run through
`Block.get_or_create_browser_state` in **both** live and script mode (the script
path's `skyvern.run_code` calls the same `CodeBlock.execute_safe`), and that
method consulted `PERSISTENT_SESSIONS_MANAGER` **first** whenever
`context.browser_session_id` was set — returning a different, unauthenticated
browser instead of the live one the agent already bound to the run.
**Fix.** `Block.get_or_create_browser_state` now (1) prefers the live
workflow-run browser (`BROWSER_MANAGER.get_for_workflow_run`), (2) falls back to
the persistent-session lookup only when there is no live browser (e.g. a
code-only run from the debugger), (3) otherwise creates one (unchanged). It's the
shared base-class path, so `PrintPageBlock` benefits too. No flag, no schema/DB
migration, no public-API change.

```diff
-        if browser_session_id and organization_id:
+        # 1. Reuse the agent's live, authenticated browser for this workflow run.
+        browser_state = app.BROWSER_MANAGER.get_for_workflow_run(workflow_run_id)
+
+        # 2. Fall back to a persistent session only when there is no live browser.
+        if not browser_state and browser_session_id and organization_id:
             browser_state = await app.PERSISTENT_SESSIONS_MANAGER.get_browser_state(browser_session_id, organization_id)
-        else:
-            browser_state = app.BROWSER_MANAGER.get_for_workflow_run(workflow_run_id)
```

This PR is independent of #<credential-test PR> and can merge in either order.
## How Has This Been Tested?
Environment: Python 3.11, `uv sync`, base commit at the `v1.0.36` tag.
- **Unit regression test** — `tests/unit/test_code_block_authed_browser_context.py`
  locks the browser-acquisition ordering. It fails on the pristine code (the Code
  Block receives the unauthenticated persistent browser) and passes with this
  change, so the regression can't silently return.
  `uv run pytest tests/unit/test_code_block_authed_browser_context.py` → 3 passed.
- **End-to-end Playwright proof (no LLM/server needed)** —
  `uv run python scripts/validate_code_block_auth_context.py` stands up an
  auth-gated mock site and drives real Playwright contexts: a fresh context hits
  `/dashboard` → sign-in page (reproduces the bug); reusing the agent's context →
  `page.evaluate()` reads the authenticated marker. Exit 0 == PASS.
N/A — covered by the unit test and the deterministic Playwright proof script above.

### DIFF
--- a/scripts/validate_code_block_auth_context.py
+++ b/scripts/validate_code_block_auth_context.py
@@ -1,0 +1,116 @@
+"""End-to-end semantic validation for the code-block authentication fix.
+
+This does NOT require a Skyvern server, a database, or an LLM key. It stands up
+a tiny auth-gated website and reproduces, with real Playwright BrowserContexts,
+the exact behavior the Skyvern fix governs:
+
+  * "agent block"  -> logs in inside a BrowserContext (cookie stored there)
+  * "code block (BROKEN)"  -> a *fresh* BrowserContext does page.goto(/dashboard)
+                              and lands on the sign-in page (the reported bug)
+  * "code block (FIXED)"   -> *reusing the agent's* BrowserContext does
+                              page.goto(/dashboard) + page.evaluate(...) and is
+                              authenticated
+
+The Skyvern fix makes the Code Block reuse the agent's live BrowserContext
+(``BROWSER_MANAGER.get_for_workflow_run``) instead of acquiring a separate one,
+which is precisely the difference between the two cases below.
+
+Run:  uv run python scripts/validate_code_block_auth_context.py
+Exit code 0 == fix behavior validated.
+"""
+
+from __future__ import annotations
+
+import http.server
+import socketserver
+import threading
+import uuid
+
+from playwright.sync_api import sync_playwright
+
+_SESSIONS: set[str] = set()
+LOGIN_HTML = b"<html><body><h1>Sign in</h1><form>sign-in page</form></body></html>"
+
+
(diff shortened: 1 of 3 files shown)


SGudbrandsson:
Closing this — the approach here (reordering `Block.get_or_create_browser_state` to prefer the workflow-run browser) does not actually fix the issue: that method is byte-identical between the last-good v1.0.33 and v1.0.36, so it was never the regression. The real fix is to run code/print blocks against the agent's live script run-context page (the authenticated BrowserContext), which I've opened as a separate PR. Superseded.

PR #6356, opened 2026-06-03
fix: run code/print blocks against the agent's live authenticated page

SGudbrandsson:
## Description

A workflow **Code Block** (and **Print Page Block**) that runs JS/Python via the Playwright `page` does not execute in the same authenticated `BrowserContext` as the preceding AI agent blocks (login / navigation / extraction) when the workflow runs in **script-execution mode** (`run_with: code`, i.e. a cached script). After the agent blocks authenticate, the code block's `page.goto(<authenticated url>)` lands on the sign-in page, because the block runs against a *different* browser context.

**Root cause.** In script-execution mode the agent blocks act on a single live `ScriptSkyvernPage` held in the script run context (`run_initializer.setup()` → `script_run_context_manager`). But a code block (`skyvern.run_code` → `CodeBlock.execute`) re-derived its page via `get_or_create_browser_state()` → `get_working_page()` instead of using that live page, so it could end up on a different, unauthenticated `BrowserContext`.

**Fix.** Add `Block.resolve_live_page()`: when a script run is active, reuse its live page — the exact authenticated page the agent blocks are using — otherwise fall back to the workflow-run browser (live/agent mode, unchanged). Wire it into `CodeBlock` and `PrintPageBlock`. No flag, no schema/DB change, no public-API change.

**Hardening (included).** `resolve_live_page` never returns a **closed** page. If the agent's live page was closed (e.g. user code closed a tab and a persisted browser session restored the dead context), the block previously got a dead handle and raised `TargetClosedError` on its first page call. Now it recovers an open page in the same authenticated browser: live page (if open) → the context's working page → a freshly opened page in that context (inherits its cookies, so it stays authenticated).

## How Has This Been Tested?

Environment: Python 3.11, `uv sync`, base `v1.0.36`.

- **Unit tests** — `tests/unit/test_code_block_uses_script_runcontext_page.py` (5 tests): the block reuses the script run-context page (and does **not** re-derive a separate browser) when a script run is active; falls back to the workflow-run browser when there is none; a **closed** run-context page is never returned (recovers the context's open working page); when no open page remains, a fresh page is opened in the authenticated context. `uv run pytest tests/unit/test_code_block_uses_script_runcontext_page.py` → 5 passed.
- **Production verification (self-hosted, script mode)** — with `id()`-level tracing at every browser-acquisition site, a real `login → navigation → extraction → code` workflow showed: before, the code block acquired a different `BrowserContext`; after, the code block's context id is identical to the agent blocks', on the warm authenticated page, with auth cookies present. End-to-end the previously-failing run now returns the authenticated content (39 extracted items, HTTP 200). The closed-page recovery was also exercised in production: a persisted session restoring a dead tab previously raised `TargetClosedError`; with this PR the block recovered an open authenticated page and the run completed.

Trace excerpt from a self-hosted run (context ids match agent ↔ code block):

```
ScriptSkyvernPage._get_or_create_browser_state  branch=workflow_run  browser_context_id=123283744923472
block reusing script run-context page           block_label=audit_gallery  browser_context_id=123283744923472
code block page state before user code           page_url=https://admin.…/…  cookie_count=31  (auth cookies present)
```

### DIFF
--- a/skyvern/forge/sdk/workflow/models/block.py
+++ b/skyvern/forge/sdk/workflow/models/block.py
@@ -425,6 +425,44 @@ class Block(BaseModel, abc.ABC):
 
         return browser_state
 
+    async def resolve_live_page(
+        self,
+        workflow_run_id: str,
+        organization_id: str | None = None,
+        browser_session_id: str | None = None,
+    ) -> Page | None:
+        """Return the Playwright page a browser-driving block (code / print) should use.
+
+        In script-execution mode the agent blocks act on a single live
+        ``ScriptSkyvernPage`` held in the script run context. A code/print block
+        that instead re-derives a page via ``get_or_create_browser_state`` ->
+        ``get_working_page`` can end up on a DIFFERENT (unauthenticated)
+        BrowserContext, so ``page.goto(<authed url>)`` bounces to the sign-in page
+        even though every agent block was authenticated. Reuse the agent's live
+        page directly so the block shares the exact same authenticated
+        BrowserContext.
+
+        Falls back to the workflow-run browser for live/agent mode (no script run
+        context exists there).
+        """
+        # Local import avoids a module-load cycle (script_skyvern_page imports
+        # heavily from forge/app, which transitively reaches this module).
+        from skyvern.core.script_generations.script_skyvern_page import script_run_context_manager
+
+        run_ctx = script_run_context_manager.get_run_context()
+        live_page = getattr(getattr(run_ctx, "page", None), "page", None) if run_ctx is not None else None
+        if live_page is not None:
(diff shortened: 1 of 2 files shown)


SGudbrandsson:
Update: verified end-to-end on a self-hosted deployment (script-execution mode, real workflow). A `login → navigation → extraction → code` run that previously bounced to the sign-in page now executes the code block in the agent's authenticated BrowserContext and returns the authenticated content (39 extracted items, HTTP 200). The closed-page recovery was also exercised in production: a persisted session that restored a dead tab previously raised `TargetClosedError`; with this PR the block recovered an open authenticated page and the run completed.

PR #6484, opened 2026-06-10
fix(SKY-10848): match editor tour steps by element, not object identity

AronPerez:


AronPerez:
Closing empty-redundant: main cumulatively absorbed this; net diff vs main empty (precedent #6485/#6394).

PR #6485, opened 2026-06-10
feat(frontend): bulk select and actions on agents list

cindehaa:

PR #6487, opened 2026-06-10
fix(SKY-10784): scope each block's downloaded files to itself inside loops

celalzamanoglu:


AronPerez:
Closing empty-redundant: main cumulatively absorbed this; net diff vs main empty (precedent #6485/#6394).

PR #6489, opened 2026-06-10
feat(SKY-9906): make persistent browser session profile generation opt-in

celalzamanoglu:


celalzamanoglu:

Source PR: https://github.com/Skyvern-AI/skyvern-cloud/pull/12007

AronPerez:
Redundant: main already contains this content (cumulative sync). The persistent-browser-session opt-in feature is fully merged on main — the `generate_browser_profile` column (models.py), routes, schemas, repositories, and the #6489 changelog bullet are all present, and main has its own canonical migration `f3a9c2b7e1d4` for the same column. The only delta remaining on this branch after merging origin/main is a duplicate migration `f80a083db810` that adds the identical column off the same parent `4dbd183edee0`, which would create divergent alembic heads + a duplicate-column failure. Net unique/genuine diff vs main is empty. Closing per precedent #6485/#6394.

PR #6492, opened 2026-06-10
feat(onboarding): thread authenticated user through workflow create/update (SKY-10722)

AronPerez:


AronPerez:
Redundant: main already contains this content (cumulative sync). After `git merge origin/main` and conflict resolution, the resulting tree is byte-for-byte identical to main (HEAD tree == origin/main tree == 255e0555). Net diff vs main is empty — the workflow create/update authenticated-user threading (SKY-10722), the generate_browser_profile column, per-run recording clips, onboarding, and bulk-action utilities are all already on main. The prior CI failures (driver.js / @clerk/clerk-react missing) and the two Copilot review comments are moot against an empty delta. Closing per precedent #6485/#6394.

PR #6611, opened 2026-06-16
fix(SKY-10799): gate sequential runs on queue order, not a single dependency edge

celalzamanoglu:


celalzamanoglu:

Source PR: https://github.com/Skyvern-AI/skyvern-cloud/pull/11949

PR #6394, opened 2026-06-05
feat(frontend): defer blank agent creation until first save

cindehaa:

PR #6656, opened 2026-06-17
Emit copilot code opt-out analytics

AronPerez:


AronPerez:
Redundant after cumulative sync #6717: merging origin/main yields an empty net diff vs main (the copilot code-mode opt-out feature, schemas, tests, and the banned_blocks/mcp_hooks helpers all landed via #6717). Per precedent #6394, closing. The open review thread on banned_blocks.py:185 is moot — that helper is already in main exactly as #6717 landed it; re-diverging to address it would defeat the redundancy close.

PR #6715, opened 2026-06-19
fix(SKY-11219): auto-fill default Google account for Sheets blocks

AronPerez:


AronPerez:
Redundant after cumulative sync #6717: net diff vs main is empty (precedent #6394). Closing.

PR #6727, opened 2026-06-20
fix(SKY-11120): launch product tour from "Take a tour" menu regardless of experiment arm

AronPerez:


AronPerez:
Redundant: cumulative sync #6730 already landed all of this PR changes on main; net diff vs main is empty. Closing per precedent #6394.

PR #6728, opened 2026-06-20
fix(SKY-11225): preserve empty FileUploadBlock no-ops

suchintan:


AronPerez:
Redundant: cumulative sync #6730 already landed all of this PR's changes on main; after merging origin/main and taking main's CodeQL-safe regex form for tests/unit/test_copilot_run_outcome_event.py, the net diff vs main is empty. Closing per precedent #6394.

This also resolves the two CodeQL "Incomplete URL substring sanitization" alerts on this PR (lines ~206 and ~429): main already uses the CodeQL-safe `re.search(r"https://example\\.com", ...)` form rather than the `"https://example.com" in ...` substring check, so the flagged code no longer exists on the path to main.

PR #6729, opened 2026-06-20
feat(SKY-8388): add Bitwarden credential UI parity

suchintan:


AronPerez:
Redundant: cumulative sync #6730 already landed all of this PR's changes (bitwarden credential UI parity with 1Password) on main.

After `git merge origin/main`, the only residual delta vs main was `tests/unit/test_copilot_run_outcome_event.py` reintroducing the CodeQL-unsafe substring form (`"https://example.com" in reason`), which is exactly what CodeQL alerts 313/314 flagged. Main already carries the CodeQL-safe `re.search(r"https://example\.com", ...)` form, so taking main's version makes the net diff vs main empty.

Net diff vs main is empty. Closing per precedent #6394.

PR #6737, opened 2026-06-22
SKY-11284: exact-name strict re-anchor + pre-navigation role/name capture

andrewneilson:


AronPerez:
Redundant: cumulative sync #6738 already landed all of this PR's changes on main; after merging origin/main the net diff vs main is empty (HEAD tree == main tree, 120a48e7b). The strict re-anchor / role-name capture delta also already landed via #6733. Closing per precedent #6394.

PR #6739, opened 2026-06-22
SKY-11188 Order browser profile lists newest first

celalzamanoglu:


AronPerez:
empty-redundant: content already on main via cumulative sync; only delta was a CodeQL substring regression (py/incomplete-url-substring-sanitization revert in tests/unit/test_copilot_run_outcome_event.py). Took main's CodeQL-clean re.search version → net diff vs main is empty. Precedent #6394.

PR #6740, opened 2026-06-22
fix(SKY-11241): fold the feasibility gate into the turn-intent classifier

andrewneilson:


AronPerez:
Closing as empty-redundant (precedent #6394). All substantive content of this PR is already on main via cumulative upstream sync. The ONLY net diff vs origin/main is tests/unit/test_copilot_run_outcome_event.py (+2/-3), which is a REGRESSION: it reverts main's CodeQL-clean `assert re.search(r"https://example\\.com", ...)` back to a bare `"https://example.com" in ...` substring check. That reintroduces 2 high-severity CodeQL alerts (incomplete URL substring sanitization, py/incomplete-url-substring-sanitization, at lines 206 & 431) which main already fixed via re.search. Taking main's version makes the diff empty, so there is nothing left to merge. Closing rather than landing a security-gate regression.

PR #6742, opened 2026-06-22
SKY-11277: bound copilot code-authoring guardrail churn before the turn timeout

andrewneilson:


andrewneilson:
Closing as superseded. SKY-11277 is already merged in cloud main (#12431), and cloud→OSS syncs are cumulative, so the next sync PR will carry these changes onto OSS main cleanly. This standalone branch only conflicts now because OSS main advanced past where it was cut (6+ sibling sync PRs landed since). Closing per the #6737 precedent rather than hand-resolving a conflict the next sync resolves for free.

PR #6744, opened 2026-06-22
SKY-11293: bound credential-priority code-authoring churn before the turn timeout

andrewneilson:


AronPerez:
Closing as empty-redundant: cumulative upstream sync already landed this content on main (net diff vs main empty after merge — even the turn_halt blocker-terminal helpers and copilot deltas this PR carried are already present on main, with the merge only re-introducing duplicates). Precedent #6394.

PR #6833, opened 2026-06-26
[codex] Add custom OSS LLM configuration

suchintan:
## Description

Adds open-source-only custom LLM configuration so local OSS users can define multiple Ollama, OpenRouter, or OpenAI-compatible models from Settings and use them through the existing model picker.

This includes backend CRUD for local custom LLM configs, LiteLLM registry wiring, TaskV2 selected-model routing, and frontend Settings UI for adding/editing/removing multiple custom LLMs.

## How Has This Been Tested?

- `uv run --extra local pytest tests/unit/test_custom_llms.py tests/unit/test_runs_v2_route.py tests/unit_tests/test_openrouter_integration.py`
- `npm run build`
- `npm run lint`
- `pre-commit run --all-files` (source hooks passed; `vitest` timed out once in an existing workflow editor perf benchmark, then `pre-commit run vitest-type-check --all-files` passed on rerun)
- Smoke tested locally with the backend, frontend, and an Ollama-compatible fake server because the local `ollama` binary is not installed. The smoke created two custom LLMs, submitted a `skyvern-2.0` task using the custom Ollama model, and run `tsk_v2_544431321573027264` completed with output `{ "smoke_result": "ok" }`. The fake Ollama request log confirmed TaskV2 LLM calls used the configured custom model `smoke-llama`.

## Artifacts

The committed PNG files were removed from the branch after review, but the screenshots are embedded here from the original PR commit.

### Custom LLM Settings

Custom LLM settings initial

### Filled Custom LLM Form

Custom LLM settings filled form

### Multiple Custom LLMs Added

Custom LLM settings after add


skyvern/forge/sdk/routes/custom_llms.py
```diff
@@ -0,0 +1,147 @@
...
+        token_type=OrganizationAuthTokenType.custom_llm,
+    )
+    custom_llms = []
+    for token in tokens:
+        try:
+            custom_llm = custom_llm_from_org_auth_token(token)
+        except Exception:
+            LOG.warning("Skipping invalid custom LLM token", token_id=token.id, exc_info=True)
+            continue
+        register_custom_llm_config(custom_llm.id, custom_llm.config)
```
andrewneilson: `list_custom_llms` is a GET but re-registers every config into the process-global `LLMConfigRegistry`; create/update and the startup loader already register, so this read-time side effect is droppable.

skyvern/forge/sdk/routes/custom_llms.py
```diff
@@ -0,0 +1,147 @@
...
+    "/custom-llms/",
+    response_model=CustomLLMResponse,
+    include_in_schema=False,
+)
+async def create_custom_llm(
+    request: CustomLLMCreateRequest,
+    current_org: Organization = Depends(org_auth_service.get_current_org),
+) -> CustomLLMResponse:
+    _require_local_custom_llms()
+    token = await app.DATABASE.organizations.create_org_auth_token(
```
andrewneilson, line 108: The `api_key` is persisted as plaintext JSON here (no `encrypted_method`) and echoed back unmasked in the list/get responses — consistent with the existing local-only posture, but worth routing through the `encrypted_method` path the repo already supports (and masking on read) if this ever graduates beyond `ENV==local`.

andrewneilson (commented):
No blockers — clean, well-tested, and correctly gated to `ENV==local`, so the cloud deployment never exposes these routes. A few pre-merge / hardening notes:

- Remove the committed `tmp/pr-screenshots/*.png` (~900KB) before merge
- `GET /custom-llms` mutates the process-global registry on read (side effect)
- `api_key` stored/echoed in plaintext; `encrypted_method` path exists if this leaves local
- `load_…(database: object)` + `getattr` — `app.DATABASE` has a concrete type
- N/A by design: localhost `api_base` (Ollama) is intended → no SSRF guard; `token_type` is a String column → no migration

Confirmed independently with a second model (Codex): no blocking correctness/security issues, and plaintext storage is consistent with the adjacent local-only settings integrations.

skyvern/forge/sdk/routes/custom_llms.py
suchintan: Fixed in 87a4f3a: list now only deserializes/masks response data and no longer re-registers configs into the process-global registry. Added a regression test asserting GET /custom-llms does not register configs.

skyvern/forge/sdk/routes/custom_llms.py
suchintan: Hardened this in 87a4f3a: API keys are now masked in create/list/update responses, and update preserves the stored key when the edit form submits the mask. Storage remains local-only/plaintext per the current OSS settings posture.

skyvern/forge/sdk/api/llm/custom_llm_registry.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def deregister_custom_llm_config(custom_llm_id: str) -> None:
+    from skyvern.forge.sdk.api.llm.config_registry import LLMConfigRegistry  # noqa: PLC0415
+
+    LLMConfigRegistry.deregister_config(custom_llm_key(custom_llm_id))
+    _custom_llm_configs.pop(custom_llm_id, None)
+
+
+def get_custom_llm_model_mappings() -> dict[str, dict[str, str]]:
```
andrewneilson: This mapping keys custom LLMs by token id only, and `TaskV2.llm_key` resolves them without checking `task_v2.organization_id`, so in a multi-org local instance one org can select another org's custom model and use its stored api_key. Bounded to multi-org local since the registry stays empty when ENV != local.

skyvern/forge/sdk/api/llm/custom_llm_registry.py
```diff
@@ -0,0 +1,115 @@
...
+    return {
+        custom_llm_model_name(custom_llm_id): {
+            "llm_key": custom_llm_key(custom_llm_id),
+            "label": f"{config.display_name} (Custom)",
```
andrewneilson: nit (optional): two custom LLMs with the same `display_name` produce identical labels; ModelSelector keys its reverse map by label, so one overwrites the other and the duplicate can't be selected. Add the id to the label or reject duplicate names.

skyvern/services/task_v2_service.py
```diff
@@ -96,6 +96,10 @@
 NAVIGATE_STRUCTURED_OUTPUT_MAX_CHARS = 10 * NAVIGATE_TERMINAL_OUTPUT_MAX_CHARS
 
 
+def _get_task_v2_llm_api_handler(task_v2: TaskV2) -> Any:
+    return LLMAPIHandlerFactory.get_override_llm_api_handler(task_v2.llm_key, default=app.LLM_API_HANDLER)
```
andrewneilson, line 122: This routes six previously-unconditional `app.LLM_API_HANDLER` callsites through `task_v2.llm_key`, so any task with a selected model — not only the new custom LLMs — now applies it here. Worth confirming that's intended for pre-existing non-custom selected models.

andrewneilson (approved):
Approving: clean OSS-local feature, prior findings resolved. Three non-blocking notes for the multi-org local edge and an override-scope check.

- Custom-LLM model map is process-global, not org-scoped (multi-org local only)
- Duplicate display names collide in ModelSelector's label map
- task_v2 now honors the selected model at six callsites for all models, not just custom

skyvern/forge/sdk/api/llm/custom_llm_registry.py
suchintan: Fixed in 240e4af: custom LLM registry entries now retain organization_id, /models returns only the current org custom models, Task/TaskV2 model resolution uses org-filtered mappings, and TaskV2 initialization rejects a custom/<id> that is not owned by the submitting org.

skyvern/forge/sdk/api/llm/custom_llm_registry.py
suchintan: Fixed in 240e4af: custom model labels now include the custom token id, so duplicate display names remain distinct in ModelSelector. Added route coverage for duplicate display names.

skyvern/services/task_v2_service.py
suchintan: Confirmed and covered in 240e4af: selected-model routing is intentional for non-custom models too. Added regression coverage that TaskV2 uses the selected non-custom model override, while custom models now get the org-ownership guard above.

suchintan:
Closing this OSS-only PR because we are moving the custom LLM configuration work into Skyvern Cloud instead.

PR #6961, opened 2026-07-01
fix: determine recording Content-Type from file extension

### LINKED ISSUE #6146: Skyvern bug in v1.0.36
Recording: files exist with .webm extension, not .mp4

The recording endpoint has a bug — it hardcodes video/mp4 but files are .webm:
javascript// In artifactServer.js:
"Content-Type": "video/mp4",  // ← wrong, files are .webm

salarkhannn:
Recording files use the `.webm` extension but the artifact server hardcoded `video/mp4` as the Content-Type for all recordings. This caused browsers to misinterpret the video format.

Instead of assuming a single format, derive the Content-Type from the file extension — matching the approach used in `agent_protocol.py` (`_VIDEO_CONTENT_TYPES_BY_EXTENSION`).

Fixes #6146

### DIFF
--- a/skyvern-frontend/artifactServer.js
+++ b/skyvern-frontend/artifactServer.js
@@ -41,7 +41,7 @@ app.get("/artifact/recording", (req, res) => {
     "Content-Range": `bytes ${start}-${end}/${videoSize}`,
     "Accept-Ranges": "bytes",
     "Content-Length": contentLength,
-    "Content-Type": "video/mp4",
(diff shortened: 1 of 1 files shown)


wintonzheng:
@salarkhannn is it breaking any feature you're using?

salarkhannn:
@wintonzheng Verified locally. .mp4 still returns video/mp4, and .webm now correctly returns video/webm, so the recordings play back in-browser instead of failing. Nothing I use breaks.

PR #6983, opened 2026-07-01
refactor(SKY-11650): remove redundant studio agent/block settings sidebar

celalzamanoglu:


AronPerez:
Empty-redundant: frontend content landed via #6984 (byte-identical studio-editor changes; #6984 was a superset with only the changelog added). merge-tree against current main is conflict-free with an empty net diff. Closing per precedent #6394.

PR #7003, opened 2026-07-02
SKY-11678 Add local credential vault persistence to OSS Docker

suchintan:
## Description

Adds a built-in encrypted Skyvern credential vault for local/self-hosted OSS deployments and wires Docker Compose to persist downloads, browser sessions, and credential vault data under `/data`. The shared app default stays `bitwarden`; OSS Compose opts into `skyvern` with `ENABLE_LOCAL_CREDENTIAL_VAULT=true`, and non-local/cloud environments stay guarded unless explicitly enabled.

Also installs Tesseract OCR plus the configured language packs in the OSS Docker image so the PDF fill block can run OCR, and updates OpenAPI/Fern/Python/TS credential vault type surfaces to include `skyvern`.

Linear: SKY-11678

## How Has This Been Tested?

- `uv run --extra server --group dev pytest -q tests/unit/test_skyvern_credential_vault_service.py tests/unit/test_credential_vault_type_override.py tests/unit/test_api_docs_taxonomy.py` -> 36 passed
- `uv run --extra server --group dev pytest -q tests/unit/test_pdf_fill_block.py` -> 26 passed
- Earlier focused credential + PDF regression suite -> 37 passed
- `pre-commit run --all-files` -> passed before final OpenAPI/.dockerignore cleanup
- `pre-commit run --files <changed files>` -> passed
- `pre-commit run` -> passed
- `docker compose config --quiet` -> passed
- `git diff --check` -> passed
- Standalone Tesseract apt smoke on `python:3.11-slim-bookworm` -> `tesseract 5.3.0`, languages `eng+spa+fra+deu+ita+por+nld+pol+rus+ukr+ara+chi_sim+chi_tra+jpn+kor+hin`
- Full image build: `docker build -t skyvern-oss-credential-vault-smoke:latest .` -> passed, image `sha256:ae8948eb59a53fe27494824a6fd4ed1ec83989fb97f0b1cec5f0f1585dffcc6a`; Tesseract layer printed `tesseract 5.3.0`. BuildKit emitted a heuristic warning for `LOCAL_CREDENTIAL_VAULT_PATH`; this env var is a path, not key material.

Self-review note: `codex review --uncommitted` was run and interrupted after the useful checks because it never emitted a final findings block. It surfaced stale OpenAPI enum docs and the missing downloads dockerignore entry; both were fixed and re-tested.

Backend/Docker change; no UI screenshots. Smoke artifact is the local image `skyvern-oss-credential-vault-smoke:latest`.


skyvern/forge/sdk/services/credential/skyvern_credential_vault_service.py
```diff
@@ -0,0 +1,169 @@
...
+            payload = self._get_fernet().decrypt(encrypted_payload).decode("utf-8")
+            item = CredentialItem.model_validate(json.loads(payload))
+        except InvalidToken as exc:
+            raise HTTPException(status_code=500, detail="Credential vault item could not be decrypted") from exc
+        except Exception as exc:
+            LOG.warning(
+                "Failed to read local credential vault item",
+                credential_id=db_credential.credential_id,
+                item_id=db_credential.item_id,
+                exc_info=True,
```
marcmuon: nit (optional): this broad `except` runs after a successful decrypt, and `exc_info=True` on a Pydantic `ValidationError` can echo decrypted field values into the traceback/logs. Dropping `exc_info` here (the `InvalidToken` branch already omits it) closes the one plaintext-to-logs vector the security lane found.

skyvern/forge/sdk/services/credential/skyvern_credential_vault_service.py
```diff
@@ -0,0 +1,169 @@
...
+
+    async def _store_item(self, item: CredentialItem) -> None:
+        item_path = self._item_path(item.item_id)
+        payload = json.dumps(item.model_dump(mode="json"), separators=(",", ":"))
+        encrypted_payload = self._get_fernet().encrypt(payload.encode("utf-8"))
+
+        temp_path = item_path.with_suffix(".tmp")
+        temp_path.write_bytes(encrypted_payload)
+        self._chmod_best_effort(temp_path, 0o600)
+        temp_path.replace(item_path)
```
marcmuon: nit (optional): the temp+`replace` dance buys atomicity but not durability without an `os.fsync(fd)` before the rename — a crash in the window can leave a zero-length item while the DB row exists (permanent 500 for that credential).

skyvern/forge/sdk/services/credential/skyvern_credential_vault_service.py
```diff
@@ -0,0 +1,169 @@
...
+        if key:
+            self._fernet = Fernet(key.encode("utf-8"))
+            return self._fernet
+
+        vault_dir = self._vault_dir()
+        key_path = vault_dir / ".fernet_key"
+        if key_path.exists():
+            key_bytes = key_path.read_bytes().strip()
+        else:
+            key_bytes = Fernet.generate_key()
```
marcmuon: nit (optional): key-file generation is last-writer-wins — two workers racing here each generate different keys and one worker's already-encrypted items become undecryptable. `os.O_CREAT|os.O_EXCL` (or documenting the single-writer assumption) closes it. Unreachable in single-container compose today.

marcmuon (approved):
Human-blessed review — three independent lanes (code-reviewer: APPROVE · architect: WATCH, nothing blocking · security: PASS), synthesized and approved by Marc.

Head reviewed: 73ca96359dd86e183b699ef4bce1283d8dd8d32a

Verified end-to-end by multiple lanes independently:
- Cloud/shared deployments stay on Bitwarden **by construction**: `is_local_credential_vault_enabled()` defaults to `not is_cloud_environment()`, cloud instantiates the service as `None`, the route double-guards, and the gating matrix is test-pinned. Clients cannot force the local vault via the API (400, tested).
- Existing self-hosters aren't flipped on upgrade (`.env` copy is guarded by `[ ! -f .env ]`; compose var is overridable). Reads always route on the **stored** `credential.vault_type`, so mixed-vault states resolve correctly and disabled vaults fail closed — never silent fallback.
- Crypto: Fernet (authenticated), random IV, atomic temp+replace, path-traversal blocked and tested, wrong key fails closed, **no hardcoded or example key anywhere** — the default that matters most at self-hoster scale, and it's right.
- Generated clients are a genuine regen (TS union gained both `skyvern` and the previously-missing `gcp`).

Hardening notes (non-blocking, the one theme all three lanes raised):
- **Key/ciphertext co-location by default**: with `LOCAL_CREDENTIAL_VAULT_KEY` unset, the auto-generated `.fernet_key` lives in the same bind-mounted dir as the encrypted items — a backup or host-dir exposure captures both. Worth a prominent note in `.env.example`/docs steering real deployments to set the key externally, and/or a startup warning when the key is co-located.
- Persisted `browser_sessions/` and `downloads/` now retain plaintext cookies/session tokens on the host across restarts — worth one line in the OSS Docker docs so operators exclude them from casual backups.
- Key rotation: rotating `LOCAL_CREDENTIAL_VAULT_KEY` bricks existing items (cached single Fernet); `MultiFernet` would make rotation survivable — fine as a follow-up.

nit (optional, not anchorable — unchanged line): the list-route `vault_type` query-param description (`routes/credentials.py` ~L1937) still reads "(e.g. 'custom', 'bitwarden', 'azure_vault')" — adding `'skyvern'` there prevents the next OpenAPI regen from reverting the doc updates in this PR.

marcmuon:
Second-pass cross-model check (codex @ xhigh, after my approval — both findings are non-blocking follow-ups; the approval stands):

**1. Orphaned vault secrets on failed test-logins** — `skyvern/forge/sdk/services/credential/skyvern_credential_vault_service.py:43`
The `/credentials/test-login` cleanup paths call `app.DATABASE.credentials.delete_credential` directly rather than dispatching through the vault service. With the local vault selected, `create_credential()` has already written the encrypted `.bin` item — so a failed/canceled/timed-out test-login soft-deletes the DB row while the secret file stays on disk indefinitely. Since the default deployment co-locates the Fernet key with the items, "deleted" credentials remain recoverable at rest. Routing those cleanup paths through the vault service (or its `post_delete_credential_item` hook) closes it.

**2. Test replaces a global singleton without restoring it** — `tests/unit/test_skyvern_credential_vault_service.py:109`
The assignment swaps `skyvern.forge.app.DATABASE.credentials` for the fake repo and never restores it — randomized test order or a mid-file failure leaves later tests seeing `_FakeCredentialRepository`. `monkeypatch.setattr(app.DATABASE, "credentials", repository)` (auto-restoring) fixes it.

PR #7029, opened 2026-07-02
feat: support browser session startup URL

### LINKED ISSUE #7028: Support startup URL when creating browser sessions
## Description

I would like to create a browser session that opens directly on a requested website.

Today, `POST /v1/browser_sessions` can create an interactive browser session, but there is no documented `url` / startup URL field in `CreateBrowserSessionRequest`. If I pass a `url` field, the session is created, but the browser still opens at `about:blank`.

This is a useful capability for integrations that launch a Skyvern browser session so a user can manually log in or set up a reusable browser profile.

## Current behavior

Calling:

```bash
curl -X POST "http://localhost:8000/v1/browser_sessions" \
  -H "Content-Type: application/json" \
  -H "x-api-key: $SKYVERN_API_KEY" \
  -d '{
    "url": "https://web.telegram.org",
    "timeout": 120,
    "generate_browser_profile": true
  }'
```
creates a browser session, but the streamed browser opens at:
```
about:blank
```

## Proposed behavior

`CreateBrowserSessionRequest` should support an optional startup URL, for example:

```
{
  "url": "https://web.telegram.org",
  "timeout": 120,
  "generate_browser_profile": true
}
```

When `url` is provided, the browser session should open directly at that URL.
When `url` is omitted, the current behavior should remain unchanged.

## Why this matters

This would make browser sessions easier to use for login/profile setup flows.
For example, an app could:

1. Create a Skyvern browser session for `https://web.telegram.org` or `https://web.whatsapp.com`.
2. Show the streamed browser to the user.
3. Let the user log in manually.
4. Save the resulting browser profile for future workflow runs.

Without startup URL support, callers need a separate workaround to navigate the browser after creating the session.

## Environment

- Skyvern: self-hosted / local Docker Compose
- Browser streaming mode: CDP
- Browser type: Chromium headful
- OS: macOS

## Artifact

https://github.com/user-attachments/assets/692846de-8714-460e-a4c7-30829d913cc7

sardortaylakov:
## Description

Resolves https://github.com/Skyvern-AI/skyvern/issues/7028

Adds optional startup URL support when creating a standalone browser session via `POST /v1/browser_sessions`.

This PR:
- Adds an optional `url` field to `CreateBrowserSessionRequest`
- Validates the URL using Skyvern's existing URL validation path
- Passes the URL into `PERSISTENT_SESSIONS_MANAGER.create_session(...)`
- Updates OpenAPI specs, Python SDK, TypeScript SDK request type, and browser-session docs
- Adds tests for accepting valid URLs, rejecting invalid URLs, and forwarding the URL to the session manager

## How Has This Been Tested?

Tested locally on macOS with Docker Compose.

Ran:

```bash
uv run pre-commit run --all-files
uv run --extra server --group dev pytest tests/unit/test_browser_session_profile_export_gate.py tests/unit/test_api_docs_taxonomy.py -q
docker compose exec -T skyvern bash -lc './run_alembic_check.sh'
npm run lint
npm run build
```
Also manually verified with Postman that creating a browser session with:
```
{
  "url": "https://web.telegram.org/a/",
  "generate_browser_profile": true,
  "timeout": 60
}
```
opens the browser session on the requested URL instead of `about:blank`.

## Artifacts

https://github.com/user-attachments/assets/b3259386-ecf5-4369-9678-41f76c3c0ba7


skyvern/client/client.py
```diff
@@ -1825,6 +1825,7 @@ def get_browser_sessions(
     def create_browser_session(
         self,
         *,
+        url: typing.Optional[str] = OMIT,
```
AronPerez: Very nice

skyvern/schemas/browser_sessions.py
```diff
@@ -7,13 +7,26 @@
...
 
 class CreateBrowserSessionRequest(BaseModel):
+    url: str | None = Field(
+        default=None,
+        description="Optional URL to open when the standalone browser session starts.",
+    )
+
+    @field_validator("url")
+    @classmethod
+    def validate_start_url(cls, value: str | None) -> str | None:
```
AronPerez, line 22: We should probably validate the url follows a known structure, what are your thoughts here, why or why not?

AronPerez (commented):
Wonderful pr, 1 question

skyvern/schemas/browser_sessions.py
sardortaylakov: @AronPerez thank you for the review 🙏  Agreed. I added validation through Skyvern’s existing `validate_url` helper so this follows the same behavior as other URL-bearing inputs.
```
from skyvern.utils.url_validators import validate_url
```

AronPerez (approved):

PR #7051, opened 2026-07-03
Show agent-scoped analytics snapshot above the runs list

AronPerez:


AronPerez:
Heads-up for visibility — auto-merge on this PR is currently gated by two verified issues, both from the file-handling batch it carries (`skyvern/forge/sdk/api/files.py`, `skyvern/forge/sdk/workflow/models/pdf_fill_block.py`, `skyvern/webeye/actions/handler_utils.py`). The same batch blocks the sibling sync PRs #7049 / #7050 / #7051.

**1. Run tests and pre-commit hooks — 3 smoke tests fail deterministically:**
- `tests/smoke_tests/test_pdf_fill_block.py::test_acroform_fill_end_to_end_through_converter`
- `tests/smoke_tests/test_pdf_fill_block.py::test_flat_fill_end_to_end_with_fake_tesseract`
- `tests/smoke_tests/test_pdf_fill_block.py::test_file_url_unwraps_upstream_block_output`

all `assert result.success is True` → `False`. These pass on `main` and on the last merged sync PR (#7048), so the regression is introduced by this batch rather than environmental.

**2. CodeQL — `[high] py/path-injection` at `skyvern/webeye/actions/handler_utils.py:30`.** The new local-file branch routes through `validate_local_file_path()`, which canonicalizes via `os.path.realpath()` and enforces containment under `DOWNLOAD_PATH/<run_id>` (raising `PermissionError` outside) — a standard path-traversal guard — so this appears to be a false positive that CodeQL isn't modeling as a sanitizer. It needs a security-owner to dismiss (or a CodeQL sanitizer annotation).

Flagging so the batch can be fixed / the alert triaged; auto-merge will proceed once both clear.

PR #7055, opened 2026-07-03
SKY-11721: enforce Copilot completion evidence provenance

andrewneilson:


AronPerez:
Heads-up for visibility — auto-merge on this PR is gated by two issues:

**1. Run tests and pre-commit hooks → `alembic check` fails:** "Your database models are out of sync with the migrations." This branch carries model changes without the corresponding migration. Per the check's own guidance, a migration needs to be generated for this branch (`alembic revision --autogenerate -m "..."`) and committed — typically done in the originating (cloud) PR so it syncs down.

**2. CodeQL — `[high] py/path-injection` at `skyvern/webeye/actions/handler_utils.py:30`.** This PR also carries the shared file-handling batch (`handler_utils.py` / `pdf_fill_block.py` / `files.py`) currently open across sibling sync PRs #7049–#7053. That alert routes through `validate_local_file_path()`, a `realpath()`-canonicalization + containment guard, so it appears to be a false positive CodeQL isn't modeling as a sanitizer — needs a security-owner to dismiss.

Flagging for visibility; auto-merge will proceed once both clear.

PR #7056, opened 2026-07-03
feat(SKY-11764): workflow studio launch polish — button grammar, pane UX, run-view defaults, resilience

celalzamanoglu:


AronPerez:
Heads-up for visibility — auto-merge on this PR is gated by two issues:

**1. Run tests and pre-commit hooks → `alembic check` fails:** "New upgrade operations detected" — the branch's models are out of sync with the migrations. The cumulative sync diff carries model changes without a matching migration (detected ops: new `workflow_run_credential_selections` table + `credential_parameters.credential_ids` / `selection_strategy` columns). Per the check's own guidance, a migration needs to be generated and committed (`alembic revision --autogenerate -m "..."`) — typically in the originating (cloud) PR so it syncs down. Note: the `alembic check` step short-circuits before pytest, so no test-level results are surfaced on this branch yet.

**2. CodeQL — `[high] py/path-injection`** in the shared file-handling batch (`handler_utils.py` / `pdf_fill_block.py` / `files.py`), currently open across sibling sync PRs #7049–#7055. That alert routes through `validate_local_file_path()`, a `realpath()`-canonicalization + containment guard, so it appears to be a false positive CodeQL isn't modeling as a sanitizer — needs a security-owner to dismiss.

Flagging for visibility; auto-merge will proceed once both clear.

PR #7058, opened 2026-07-03
[frontend] Fix run/block duration display dropping days past 24h

### LINKED ISSUE #7057: Run/block duration display drops a full day past 24h (hours wrap at 24)
### What happened

Run and block duration labels in the UI wrap the hour component at 24, so any run that lasts a day or longer shows the wrong time. A run that actually took 25 hours is displayed as `1h 0m 0s`.

### Where

`skyvern-frontend/src/routes/workflows/utils.ts` — `toDuration()`:

```ts
export const toDuration = (seconds: number): Duration => {
  let minutes = Math.floor(seconds / 60);
  let hours = Math.floor(minutes / 60);
  seconds = seconds % 60;
  minutes = minutes % 60;
  hours = hours % 24;   // <- drops whole days
  ...
```

The returned `Duration` type only has `hour`/`minute`/`second`, and `formatDuration()` only renders those three, so there is no day field to carry the overflow — the `% 24` just discards it. (The related `formatMs()` in `src/util/utils.ts` does the same `% 24` but is fine there because it reports `days` separately.)

This is reachable from normal run views, e.g. `getRunDurationLabel()` in `recentActivity/runActivity.ts` and the block duration labels in `WorkflowRunTimelineBlockItem.tsx`, both of which feed real elapsed seconds (`finished_at - started_at`) into `toDuration`.

### Steps to reproduce

- Open a workflow run (or block) whose elapsed time is ≥ 24h.
- The duration label reads e.g. `1h 0m 0s` for a 25h run instead of `25h 0m 0s`.

Or in isolation:

```ts
formatDuration(toDuration(90000)) // => "1h 0m 0s", expected "25h 0m 0s"
```

### Expected

The hour field should carry the total hours (`25h 0m 0s`), since there is no day component in the label.

I have a small fix + tests ready and will open a PR referencing this issue.

abhay-codes07:
Fixes #7057

### Problem

`toDuration()` in `skyvern-frontend/src/routes/workflows/utils.ts` reduced the hour component with `hours % 24`, but the `Duration` type it returns has no day field and `formatDuration()` only renders hours/minutes/seconds. So any run or block that lasts 24h+ silently loses a full day in its label — a 25h run shows `1h 0m 0s`.

This is reachable from normal run views: `getRunDurationLabel()` (`recentActivity/runActivity.ts`) and the block duration labels in `WorkflowRunTimelineBlockItem.tsx` both feed real elapsed seconds (`finished_at - started_at`) into `toDuration`.

### Fix

Drop the `hours % 24` so the hour field carries the total hours. Minimal, no interface or display-format change — hours can now simply exceed 24.

```
formatDuration(toDuration(90000)) // before: "1h 0m 0s"  after: "25h 0m 0s"
```

### Tests

Added tests to `utils.test.ts` covering sub-minute, minute/second, hour/minute/second, fractional-second flooring, and the >24h / multi-day regression cases. Confirmed the new duration tests fail on the old `% 24` code and pass with the fix; full file suite is green (`vitest run src/routes/workflows/utils.test.ts` → 20 passed). Prettier and ESLint clean on the changed files.

### DIFF
--- a/skyvern-frontend/src/routes/workflows/utils.ts
+++ b/skyvern-frontend/src/routes/workflows/utils.ts
@@ -169,10 +169,9 @@ export interface Duration {
 
 export const toDuration = (seconds: number): Duration => {
   let minutes = Math.floor(seconds / 60);
-  let hours = Math.floor(minutes / 60);
+  const hours = Math.floor(minutes / 60);
   seconds = seconds % 60;
   minutes = minutes % 60;
-  hours = hours % 24;
 
   return {
     hour: Math.floor(hours),
--- a/skyvern-frontend/src/routes/workflows/utils.test.ts
+++ b/skyvern-frontend/src/routes/workflows/utils.test.ts
@@ -1,8 +1,10 @@
 import { describe, expect, it } from "vitest";
 
 import {
+  formatDuration,
   normalizeJsonParameterFormValue,
   parseJsonWorkflowParameterValue,
+  toDuration,
   validateJsonWorkflowParameterValue,
 } from "./utils";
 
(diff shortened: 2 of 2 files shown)


abhay-codes07:
Hey @suchintan @wintonzheng a small one when either of you gets a sec. Noticed run/block duration labels wrap the hours at 24, so a run over a day shows something like `1h 0m 0s` for 25h. One-line fix plus tests here. Happy to adjust if you'd rather show a `Nd` days prefix instead ,went with total hours to keep the change minimal. Thanks!

skyvern-frontend/src/routes/workflows/utils.ts
```diff
@@ -169,10 +169,9 @@ export interface Duration {
 
 export const toDuration = (seconds: number): Duration => {
   let minutes = Math.floor(seconds / 60);
-  let hours = Math.floor(minutes / 60);
+  const hours = Math.floor(minutes / 60);
```
AronPerez, line 172: now we get 25h, very nice

AronPerez (approved):

PR #7063, opened 2026-07-03
[frontend] Substitute all parameter references when building cache key value

### LINKED ISSUE #7060: Cache key value is dropped when a cache key template references the same parameter more than once
### What happened

`constructCacheKeyValueFromParameters` in `skyvern-frontend/src/routes/workflows/editor/utils.ts` fills in a cache key template by replacing each `{{name}}` with its parameter value:

```ts
for (const [name, value] of Object.entries(parameters)) {
  if (value === null || value === undefined || value === "") continue;
  codeKey = codeKey.replace(`{{${name}}}`, value.toString());
}
if (codeKey.includes("{") || codeKey.includes("}")) return "";
```

`String.prototype.replace` with a string argument only replaces the first match. If a cache key template references the same parameter more than once, the later placeholders survive, the `{`/`}` guard then treats the key as unresolved, and the function returns an empty string. The cache lookup for that run is silently disabled.

### Reproduce

```ts
constructCacheKeyValueFromParameters({ codeKey: "{{a}}/{{a}}", parameters: { a: "x" } })
// current:  ""
// expected: "x/x"
```

Templates that reference distinct parameters, each once, are unaffected.

### Expected

All occurrences of a parameter should be substituted, e.g. with `replaceAll`.

abhay-codes07:
Fixes #7060

### Problem

`constructCacheKeyValueFromParameters` (`skyvern-frontend/src/routes/workflows/editor/utils.ts`) substituted parameters with `String.replace`, which replaces only the first match. A cache key template that references the same parameter more than once (for example `{{a}}/{{a}}`) kept the second placeholder, so the trailing `{`/`}` guard returned an empty string and the cache lookup for that run was silently disabled.

### Fix

Use `replaceAll` so every occurrence is substituted.

```
constructCacheKeyValueFromParameters({ codeKey: "{{a}}/{{a}}", parameters: { a: "x" } })
// before: ""
// after:  "x/x"
```

Templates with distinct parameters, each referenced once, are unaffected.

### Tests

Extended `editor/utils.test.ts` with cases for a single reference, distinct parameters, the repeated reference regression, an unresolved placeholder returning `""`, and skipped null/undefined/empty values. Confirmed the repeated-reference test fails on the old code and passes with the fix. `vitest run` on the file is green, prettier and eslint clean.

### DIFF
--- a/skyvern-frontend/src/routes/workflows/editor/utils.ts
+++ b/skyvern-frontend/src/routes/workflows/editor/utils.ts
@@ -191,7 +191,7 @@ const constructCacheKeyValueFromParameters = (opts: {
       continue;
     }
 
-    codeKey = codeKey.replace(`{{${name}}}`, value.toString());
+    codeKey = codeKey.replaceAll(`{{${name}}}`, value.toString());
   }
 
   if (codeKey.includes("{") || codeKey.includes("}")) {
--- a/skyvern-frontend/src/routes/workflows/editor/utils.test.ts
+++ b/skyvern-frontend/src/routes/workflows/editor/utils.test.ts
@@ -5,6 +5,7 @@ import {
   WorkflowParameterValueType,
 } from "../types/workflowTypes";
 import {
+  constructCacheKeyValueFromParameters,
   getInitialParameters,
   skyvernCredentialToParameterYAML,
 } from "./utils";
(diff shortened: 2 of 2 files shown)


abhay-codes07:
@suchintan @wintonzheng quick one when you have a moment. Cache key construction uses String.replace, which only swaps the first placeholder, so a key template that reuses the same parameter twice comes back empty and the run skips its cache lookup. Switched to replaceAll and added tests. Appreciate it!
cc @AronPerez

skyvern-frontend/src/routes/workflows/editor/utils.ts
```diff
@@ -191,7 +191,7 @@ const constructCacheKeyValueFromParameters = (opts: {
       continue;
     }
 
-    codeKey = codeKey.replace(`{{${name}}}`, value.toString());
+    codeKey = codeKey.replaceAll(`{{${name}}}`, value.toString());
```
AronPerez, line 194: This literal-string `replaceAll` safely replaces repeated placeholders without regex escaping concerns, while the unchanged guard below still rejects genuinely unresolved parameters and the regression test covers `{{a}}/{{a}}`.

AronPerez (approved):

Issue #242, opened 2024-04-29
How does skyvern integrate with ollama litellm


alexsiu398:
Is there a way or tutorial on how to configure ollama litellm to work with skyvern? How can skyvern work with a local llm?

suchintan:
Here's an example where @ykeremy built out bedrock support within Skyvern

https://github.com/Skyvern-AI/skyvern/pull/251/files

Are you open to opening a PR for ollama + litellm? We'd love a contribution here!

suchintan:
Ignore the files in the experimentation module. The other configs are all you need!

santiagoblanco22:
Nice. Now GPT4 60 USD in 3 days. :( Ollama is awesome! I don't know how to help!

suchintan:
@santiagoblanco22 we would love a contribution here!! Or maybe we can ask for people's help in our discord?

GPT4 is super expensive. Try it with Claude 3 sonnet instead

OB42:
hi, I'm currently trying to add it. :)

Do you think we should allow all ollama models? should we ask the user for a specific model name(as a string)? or a numbered choice like for anthropic with just llama3/mistral maybe llava?

OB42:
FYI for now it seems that most models available on Ollama are not good enough weak for Skyvern , at least on my computer, so it seems pointless to add models that would not work well.

Maybe it could work with a 34/70B model with no quantization, but you would need a very beefy setup, at that point you'd probably be better off using bedrock or anthropic IMO

HarryBak:
> FYI for now it seems that most models available on Ollama are not good enough for Skyvern , at least on my computer, so it seems pointless to add models that would not work well.
> 
> Maybe it could work with a 34/70B model with no quantization, but you would need a very beefy setup, at that point you'd probably be better off using bedrock/anthropic IMO


What models did you try using?

At what stage was it getting stuck?

Any other information you could share that may help the research into making this reality?

OB42:
> > FYI for now it seems that most models available on Ollama are not good enough for Skyvern , at least on my computer, so it seems pointless to add models that would not work well.
> > Maybe it could work with a 34/70B model with no quantization, but you would need a very beefy setup, at that point you'd probably be better off using bedrock/anthropic IMO
> 
> What models did you try using?
> 
> At what stage was it getting stuck?
> 
> Any other information you could share that may help the research into making this reality?

If I remember correctly, LLama-2, Mistral 7B, Phi3.
[...]

PR #7065, opened 2026-07-03
fix(SKY-11826): gate the legacy Visual/Code toggle to /edit and anchor it under the header

celalzamanoglu:


AronPerez:
Heads-up: this sync mirror carries the same unmerged file-handling batch as #7049–#7056 and currently fails two of the batch blockers:

1. **Run alembic check fails** — the credential-rotation models (`workflow_run_credential_selections` table; `credential_parameters.credential_ids` / `selection_strategy` columns) have no matching migration in the mirror, so the test gate short-circuits before pytest runs ([run 28685953228](https://github.com/Skyvern-AI/skyvern/actions/runs/28685953228)).
2. **CodeQL umbrella check fails** on the open \[high\] py/path-injection alert (#242) at `files.py:235` / `handler_utils.py:30` (`validate_local_file_path`).

The pdf_fill smoke-test regression seen on #7049–#7053 could not be evaluated here because the alembic check aborts the job first. Once the batch blockers are resolved upstream, this PR should clear via a fresh sync.

PR #7066, opened 2026-07-03
fix(SKY-11831): keep the API & Webhooks label on one line under flex squeeze

celalzamanoglu:


AronPerez:
Heads-up: this sync mirror carries the same unmerged file-handling batch as #7049–#7056/#7065 and currently fails two of the batch blockers:

1. **Run alembic check fails** — the credential-rotation models (`workflow_run_credential_selections` table; `credential_parameters.credential_ids` / `selection_strategy` columns) have no matching migration in this mirror (the migration exists on #7053's branch as `aff1632dc377` but is absent here), so the test gate short-circuits before pytest runs ([run 28687075833](https://github.com/Skyvern-AI/skyvern/actions/runs/28687075833)).
2. **CodeQL umbrella check fails** on the \[high\] py/path-injection alert at `validate_local_file_path` (`files.py` / `handler_utils.py`) — a fix for this is in flight on #7049.

Once #7049 and #7053 land, this PR should clear via update-branch / a fresh sync.

AronPerez:
Verified: merge with origin/main yields zero net diff after #7049/#7053/#7055/#7056 landed; closing per precedent #6394.

PR #7069, opened 2026-07-04
SKY-11591: author-time output contract for Copilot build-time repair

andrewneilson:


AronPerez:
Heads-up @suchintan @andrewneilson — a root cause + durable ask from the sync pipeline (not a blocker for this PR, which already merged green):

This PR's sync mirror transiently failed CodeQL (path-injection) because it was cut from an upstream snapshot that reverts main's CodeQL-clean path-sanitizer shape in `files.py`/`handler_utils.py` back to the earlier variant CodeQL's taint model doesn't recognize as a sanitizer. It was auto-resolved by restoring main's version of those files (main-wins, keeping this PR's feature delta), so it merged clean.

The durable problem: **upstream (the private repo) still carries the pre-fix sanitizer variant.** Until the public fix — `validate_local_file_path` restructured to `os.path.realpath(...)` + a `startswith(<DOWNLOAD_PATH>/<run_id>)` containment check *in the same function* (the shape CodeQL models as a sanitizer; this is what closed alert #242 as *fixed* on main) — is synced back upstream, **every future sync mirror touching these files will re-fire the CodeQL path-injection alert and need per-PR restoration.**

Could one of you sync that fix back upstream so mirrors stop regressing? Happy to point at the exact main commit if that helps.

PR #7075, opened 2026-07-04
[frontend] Fix duplicate block label on single-collision rename

### LINKED ISSUE #7074: Renaming a workflow block to collide with one other block produces a duplicate label
### What happened

`getUniqueLabelForExistingNode` in `skyvern-frontend/src/routes/workflows/editor/workflowEditorUtils.ts` is supposed to return a label that is not already used. Its loop bound is off:

```ts
if (!existingLabels.includes(label)) {
  return label;
}
for (let i = 2; i < existingLabels.length + 1; i++) {
  const candidate = `${label}_${i}`;
  if (!existingLabels.includes(candidate)) {
    return candidate;
  }
}
return label; // falls through to the colliding label
```

With one existing label the loop range is `2 <= i < 2`, so it never runs and the function returns the colliding label unchanged. More generally, `N` existing labels give only `N-1` candidates (`label_2 .. label_N`), which is not always enough, so it can fall through and return a duplicate.

The caller `useLabelChangeHandler.ts` builds `existingLabels` from every other block node, then renames the edited block to the returned value. So renaming a block to match exactly one other block (the common single-collision case) creates two blocks with the same label. Block labels must be unique because output parameters are keyed by them (`<label>_output`).

### Reproduce

```ts
getUniqueLabelForExistingNode("block_1", ["block_1"]) // returns "block_1", expected "block_1_2"
getUniqueLabelForExistingNode("foo", ["foo", "foo_2"]) // returns "foo", expected "foo_3"
```

### Expected

A label that is not already present. The sibling `generateNodeLabel` uses the correct bound `existingLabels.length + 2`.

I have a one-line fix plus tests ready and will open a PR that references this issue.

abhay-codes07:
Fixes #7074

### Problem

`getUniqueLabelForExistingNode` (`skyvern-frontend/src/routes/workflows/editor/workflowEditorUtils.ts`) looped `for (i = 2; i < existingLabels.length + 1; i++)`. With one existing label the loop never ran, and with N labels it tried only N-1 candidates (`label_2 .. label_N`), so it fell through and returned the colliding label unchanged.

Its caller `useLabelChangeHandler.ts` builds `existingLabels` from every other block node and renames the edited block to the returned value, so renaming a block to match exactly one other block produced two blocks with the same label. Labels must be unique because output parameters are keyed by them (`<label>_output`).

### Fix

Widen the bound to `existingLabels.length + 2`, matching the sibling `generateNodeLabel`, so a free suffix is always found.

```
getUniqueLabelForExistingNode("block_1", ["block_1"]) // before "block_1", after "block_1_2"
getUniqueLabelForExistingNode("foo", ["foo", "foo_2"]) // before "foo",     after "foo_3"
```

### Tests

Added `workflowEditorUtils.getUniqueLabelForExistingNode.test.ts` covering no collision, single collision, next-free-suffix, and a property check that the result is never already present for collision sets of size 1 through 25. Confirmed the collision tests fail on the old bound and pass with the fix. `vitest run` on the file is green, prettier and eslint clean.

### DIFF
--- a/skyvern-frontend/src/routes/workflows/editor/workflowEditorUtils.getUniqueLabelForExistingNode.test.ts
+++ b/skyvern-frontend/src/routes/workflows/editor/workflowEditorUtils.getUniqueLabelForExistingNode.test.ts
@@ -1,0 +1,36 @@
+import { describe, expect, it } from "vitest";
+
+import { getUniqueLabelForExistingNode } from "./workflowEditorUtils";
+
+describe("getUniqueLabelForExistingNode", () => {
+  it("returns the label unchanged when it does not collide", () => {
+    expect(getUniqueLabelForExistingNode("block_1", ["other"])).toBe("block_1");
+    expect(getUniqueLabelForExistingNode("block_1", [])).toBe("block_1");
+  });
+
+  it("suffixes on a single collision instead of returning the duplicate", () => {
+    // Regression: the loop bound was existingLabels.length + 1, so with one
+    // existing label the loop never ran and the colliding label was returned.
+    expect(getUniqueLabelForExistingNode("block_1", ["block_1"])).toBe(
+      "block_1_2",
+    );
+  });
+
+  it("finds the next free suffix when earlier suffixes are taken", () => {
+    expect(getUniqueLabelForExistingNode("foo", ["foo", "foo_2"])).toBe(
+      "foo_3",
+    );
+    expect(getUniqueLabelForExistingNode("x", ["x", "x_2", "x_3"])).toBe("x_4");
+  });
+
+  it("always returns a label not already present", () => {
+    for (let n = 1; n <= 25; n++) {
+      const existing = [
+        "dup",
+        ...Array.from({ length: n }, (_, i) => `dup_${i + 2}`),
+      ];
+      const result = getUniqueLabelForExistingNode("dup", existing);
+      expect(existing).not.toContain(result);
+    }
+  });
+});
--- a/skyvern-frontend/src/routes/workflows/editor/workflowEditorUtils.ts
+++ b/skyvern-frontend/src/routes/workflows/editor/workflowEditorUtils.ts
@@ -3876,7 +3876,7 @@ function getUniqueLabelForExistingNode(
   if (!existingLabels.includes(label)) {
     return label;
   }
-  for (let i = 2; i < existingLabels.length + 1; i++) {
+  for (let i = 2; i < existingLabels.length + 2; i++) {
     const candidate = `${label}_${i}`;
     if (!existingLabels.includes(candidate)) {
       return candidate;


abhay-codes07:
@suchintan @wintonzheng small correctness one when you have a sec. Renaming a block so it matches exactly one other block currently returns the same name back, so you end up with two blocks sharing a label. The unique-suffix loop was one iteration short, so it fell through on a single collision. One line to match the bound the sibling helper already uses, plus tests. Thanks!

skyvern-frontend/src/routes/workflows/editor/workflowEditorUtils.ts
```diff
@@ -3876,7 +3876,7 @@ function getUniqueLabelForExistingNode(
   if (!existingLabels.includes(label)) {
     return label;
   }
-  for (let i = 2; i < existingLabels.length + 1; i++) {
+  for (let i = 2; i < existingLabels.length + 2; i++) {
```
AronPerez, line 3879: Matching `generateNodeLabel`'s `existingLabels.length + 2` bound gives N suffix candidates for N existing labels, so the original-label collision leaves at least one candidate free and prevents fall-through.

skyvern-frontend/src/routes/workflows/editor/workflowEditorUtils.getUniqueLabelForExistingNode.test.ts
```diff
@@ -0,0 +1,36 @@
...
+describe("getUniqueLabelForExistingNode", () => {
+  it("returns the label unchanged when it does not collide", () => {
+    expect(getUniqueLabelForExistingNode("block_1", ["other"])).toBe("block_1");
+    expect(getUniqueLabelForExistingNode("block_1", [])).toBe("block_1");
+  });
+
+  it("suffixes on a single collision instead of returning the duplicate", () => {
+    // Regression: the loop bound was existingLabels.length + 1, so with one
+    // existing label the loop never ran and the colliding label was returned.
+    expect(getUniqueLabelForExistingNode("block_1", ["block_1"])).toBe(
```
AronPerez, line 14: This assertion directly covers the ticket's single-collision regression and expected `block_1_2` result.

skyvern-frontend/src/routes/workflows/editor/workflowEditorUtils.getUniqueLabelForExistingNode.test.ts
```diff
@@ -0,0 +1,36 @@
...
+      "block_1_2",
+    );
+  });
+
+  it("finds the next free suffix when earlier suffixes are taken", () => {
+    expect(getUniqueLabelForExistingNode("foo", ["foo", "foo_2"])).toBe(
```
AronPerez, line 20: This assertion covers the chained-collision case by proving the occupied `_2` suffix advances to `_3`.

AronPerez (approved):

PR #7078, opened 2026-07-04
[webeye] Parse enable_download header as a boolean instead of presence

### LINKED ISSUE #7077: enable_download extra header treats any non-empty value as true (including "false")
### What happened

`parse_extra_headers` in `skyvern/webeye/browser_factory.py` reads the internal `enable_download` header like this:

```python
elif key_lower == "enable_download":
    enable_download = bool(value)
```

`value` is the header string, and `bool(non_empty_string)` is always `True`. So `enable_download: "false"`, `"0"`, and `"no"` all resolve to `True`; only an empty string yields `False`. The value is effectively ignored and any presence enables downloads.

`enable_download` gates the `CDPDownloadInterceptor` (browser_factory.py around line 1057), so a caller that explicitly sends `enable_download: false` still gets download interception turned on.

### Reproduce

```python
from skyvern.webeye.browser_factory import parse_extra_headers
parse_extra_headers({"enable_download": "false"}).enable_download  # True, expected False
parse_extra_headers({"enable_download": "0"}).enable_download      # True, expected False
```

### Expected

The boolean text should be parsed, not the string's presence. The sibling branch in the same loop already does this for the fresh-context header: `use_fresh_context = value.lower() == "true"`.

I have a one-line fix plus tests ready and will open a PR that references this issue.

abhay-codes07:
Fixes #7077

### Problem

`parse_extra_headers` (`skyvern/webeye/browser_factory.py`) set `enable_download = bool(value)`, but `value` is a header string. `bool(non_empty_string)` is always `True`, so `enable_download: "false"`, `"0"`, and `"no"` all enabled downloads; only an empty string yielded `False`. Since `enable_download` gates the `CDPDownloadInterceptor`, a caller sending `enable_download: false` still got download interception.

### Fix

Parse the boolean text: `value.lower() == "true"`, mirroring the sibling `X-Skyvern-Fresh-Context` branch in the same loop.

```
parse_extra_headers({"enable_download": "false"}).enable_download  # before True, after False
parse_extra_headers({"enable_download": "true"}).enable_download   # True
```

### Tests

Added a `TestParseExtraHeaders` class in `tests/unit/test_browser_header_sanitization.py` covering the true values, the false/0/no/empty values, the absent default, internal-header stripping, and the fresh-context boolean. Confirmed the false-values test fails on the old code and passes with the fix. ruff check and ruff format are clean.


abhay-codes07:
@suchintan @wintonzheng small one when you get a chance. The enable_download extra header was read with bool(value), so any non-empty string, including "false", turned downloads on. Switched it to parse the boolean text the way the fresh-context header right above it already does. Tests included. Thanks!

wintonzheng (approved):

skyvern/webeye/browser_factory.py
```diff
@@ -97,7 +97,7 @@ def parse_extra_headers(extra_http_headers: dict[str, str] | None) -> ParsedBrow
             if key_lower == FRESH_CONTEXT_HEADER.lower():
                 use_fresh_context = value.lower() == "true"
             elif key_lower == "enable_download":
-                enable_download = bool(value)
```
wintonzheng, line 100: this was a hack i put in.

if you send {"enable_download": "1"} it should still work. so the "truthy" function is more accurate

skyvern/webeye/browser_factory.py
```diff
@@ -97,7 +97,7 @@ def parse_extra_headers(extra_http_headers: dict[str, str] | None) -> ParsedBrow
...
+                enable_download = value.lower() == "true"
```
AronPerez: Could you normalize surrounding whitespace before comparing (and cover it in the regression test), since `" true "` was truthy before this change but is now interpreted as `False`?

AronPerez (changes requested):

skyvern/webeye/browser_factory.py
abhay-codes07: Good point, thanks. Switched to a small truthy helper so {"enable_download": "1"} still enables downloads, along with "yes"/"y"/"on", all trimmed and case-insensitive. "false"/"0"/"no" and any other value stay off. Applied the same helper to the fresh-context flag for consistency.

skyvern/webeye/browser_factory.py
abhay-codes07: Done. Both flags now go through _header_flag_is_truthy, which strips surrounding whitespace before matching, so " true " is truthy again. Added regressions for the whitespace cases (" true ", "	true
", " 1 ", " YES ") and for the fresh-context flag, plus the "1"/"yes"/"on" truthy set and a "disabled"/whitespace-only falsy check.

abhay-codes07:
Pushed 4bd0294. Both boolean headers (enable_download and fresh-context) now parse through a truthy helper that trims whitespace and matches "true"/"1"/"yes"/"y"/"on" case-insensitively, so "1" and " true " work again while "false"/"0"/"no" stay off. Added the whitespace and truthy/falsy regressions. Local header suite is green (18 passed). Thanks both.

PR #7080, opened 2026-07-04
[webeye] Position cursor at edge coordinate 0 in drag and left_mouse

### LINKED ISSUE #7079: drag and left_mouse skip cursor positioning when a coordinate is 0 (screen edge)
### What happened

`drag` and `left_mouse` in `skyvern/webeye/actions/handler_utils.py` gate the move-to-start on a truthiness check:

```python
async def drag(page, start_x=None, start_y=None, path=None):
    if start_x and start_y:                       # 0 is falsy
        await EventStrategyFactory.move_cursor(page, start_x, start_y)
    ...

async def left_mouse(page, x, y, direction):
    if x and y:                                   # 0 is falsy
        await EventStrategyFactory.move_cursor(page, x, y)
```

A coordinate of `0` (the top or left screen edge) is falsy, so the cursor is never moved before `mouse.down()`. The drag or click then fires at wherever the cursor happened to be, not at the requested edge coordinate. `drag` even uses `is not None` for the very same `start_x`/`start_y` a few lines later (setting `last_x`/`last_y` and the final `sync_cursor_position`), which shows the intended check.

### Expected

Coordinate `0` is a valid position and should move the cursor. Use `is not None` so only a missing coordinate skips the move.

### Reproduce

Calling `drag(page, start_x=0, start_y=5)` never invokes `move_cursor`, so the drag starts from the previous cursor position instead of `(0, 5)`. Same for `left_mouse(page, 0, 0, "down")`.

abhay-codes07:
Fixes #7079

### Problem

`drag` and `left_mouse` (`skyvern/webeye/actions/handler_utils.py`) gated the move-to-start on `if start_x and start_y` / `if x and y`. Coordinate `0` is falsy, so a drag or click at a screen edge (x or y == 0) skipped the cursor move and the press fired at the previous cursor position. `drag` already uses `is not None` for the same coordinates when setting `last_x`/`last_y` and calling `sync_cursor_position`, confirming the intended check.

### Fix

Use `is not None` so `0` is treated as a real coordinate; only a missing (`None`) coordinate skips the move.

### Tests

Added `tests/unit/webeye/test_handler_utils_mouse.py` (mocked page and EventStrategyFactory) asserting `move_cursor` is invoked for zero coordinates in both functions and skipped only when a coordinate is `None`. Confirmed the zero-coordinate cases fail on the old truthiness check and pass with the fix. ruff check and ruff format are clean.

### DIFF
--- a/skyvern/webeye/actions/handler_utils.py
+++ b/skyvern/webeye/actions/handler_utils.py
@@ -114,7 +114,7 @@ async def keypress(page: Page, keys: list[str], hold: bool = False, duration: fl
 async def drag(
     page: Page, start_x: int | None = None, start_y: int | None = None, path: list[tuple[int, int]] | None = None
 ) -> None:
-    if start_x and start_y:
+    if start_x is not None and start_y is not None:
         await EventStrategyFactory.move_cursor(page, start_x, start_y)
     await page.mouse.down()
     path = path or []
@@ -131,7 +131,7 @@ async def drag(
 
 
 async def left_mouse(page: Page, x: int | None, y: int | None, direction: Literal["down", "up"]) -> None:
-    if x and y:
+    if x is not None and y is not None:
         await EventStrategyFactory.move_cursor(page, x, y)
     if direction == "down":
         await page.mouse.down()
--- a/tests/unit/webeye/test_handler_utils_mouse.py
+++ b/tests/unit/webeye/test_handler_utils_mouse.py
@@ -1,0 +1,70 @@
+"""Regression tests for cursor positioning in drag/left_mouse.
+
+Both functions gated the move-to-start on a truthiness check, so a coordinate
+of 0 (a screen edge) was treated as "no coordinate" and the cursor was never
+moved before the mouse press. The drag/click then fired at the previous cursor
+position instead of the requested edge coordinate.
+"""
+
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+from skyvern.webeye.actions import handler_utils
+
+
+def _fake_page() -> MagicMock:
+    page = MagicMock()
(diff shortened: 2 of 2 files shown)


abhay-codes07:
@suchintan @wintonzheng another quick correctness one. drag and left_mouse skip moving the cursor when a coordinate is 0 because of a truthiness check, so a drag or click at the top or left edge fires from the previous position. drag already uses is not None for the same values a couple lines down, so this just makes the guard consistent. Tests added. Appreciate it!

skyvern/webeye/actions/handler_utils.py
```diff
@@ -114,7 +114,7 @@ async def keypress(page: Page, keys: list[str], hold: bool = False, duration: fl
 async def drag(
     page: Page, start_x: int | None = None, start_y: int | None = None, path: list[tuple[int, int]] | None = None
 ) -> None:
-    if start_x and start_y:
+    if start_x is not None and start_y is not None:
```
AronPerez, line 117: The explicit checks for both coordinates correctly allow edge starts such as `(0, 5)` while still skipping the initial move if either value is `None`.

skyvern/webeye/actions/handler_utils.py
```diff
@@ -131,7 +131,7 @@ async def drag(
 
 
 async def left_mouse(page: Page, x: int | None, y: int | None, direction: Literal["down", "up"]) -> None:
-    if x and y:
+    if x is not None and y is not None:
```
AronPerez, line 134: This matching guard fixes `left_mouse` for zero-valued axes and preserves the existing no-move behavior when either coordinate is absent.

AronPerez (approved):

PR #7125, opened 2026-07-06
SKY-11830 Fix previous extracted info truncation

AronPerez:
Linear: https://linear.app/skyvern/issue/SKY-11830

## Bug
`truncate_previous_extracted_information` could corrupt under-budget dict values in `previous_extracted_information`, turning nested lists or dicts into tail-sliced strings.

## Root cause
The dict branch always called `_crop_dict`, which applies a per-key token budget even when the rendered whole value already fits `max_tokens`.

## Fix
Return the original value unchanged as soon as its rendered token count is within budget, before dispatching to string/list/dict crop behavior.

## Test
Added a regression for an under-budget dict containing a nested `results` list. The test fails on the old behavior because `results` becomes a truncated string, and passes with the shared short-circuit.

Verification:
- `uv sync --extra server --group dev`
- `uv run pytest tests/unit/test_prompt_truncation.py::test_truncate_dict_under_budget_preserves_nested_list_value` (red before fix, green after)
- `uv run pytest tests/unit/test_prompt_truncation.py`
- `uv run ruff check skyvern tests/unit/test_prompt_truncation.py`
- `uv run ruff format --check skyvern tests/unit/test_prompt_truncation.py`
- commit pre-commit hooks

### DIFF
--- a/skyvern/utils/prompt_truncation.py
+++ b/skyvern/utils/prompt_truncation.py
@@ -83,6 +83,8 @@ def truncate_previous_extracted_information(
 
     rendered_before = value if isinstance(value, str) else json.dumps(value, default=str)
     before_tokens = count_tokens(rendered_before)
+    if before_tokens <= max_tokens:
+        return value
 
     if isinstance(value, str):
         result: Any = _crop_string(value, max_tokens)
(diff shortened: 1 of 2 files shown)


AronPerez:
Closing as a duplicate. **SKY-11830** already has an open community PR #7064 that predates this one, so we're deferring to the contributor's work there — please review #7064 instead. This PR was opened by an automated agent before we spotted the existing contributor PR; apologies for the duplicate noise. 🙏

PR #7076, opened 2026-07-04
fix: stop reseeding the global random PRNG in generate_random_string

thribhuvan003:
## Description

`generate_random_string` calls `random.seed(os.urandom(16))` on every invocation, which:

- **mutates the process-wide global `random` state** — any other code relying on a seeded global PRNG (tests, reproducible sampling) gets silently clobbered whenever a random string is generated;
- reseeds the Mersenne Twister (2.5 KB state re-init) per call for a 5-character string;
- is not thread-safe: concurrent callers race on the shared global state.

This PR switches to a **module-level `random.SystemRandom()`** instance: it draws directly from OS entropy (same source as `secrets`), never touches the global PRNG, needs no seeding, and is safe under concurrency. Output format and the `RANDOM_STRING_POOL` alphabet are unchanged; the now-unused `import os` is removed.

Fixes #7072

## How Has This Been Tested?

Verified the global-state pollution and the fix with a standalone reproduction (Python 3.13, Windows):

```text
random.seed(42) → random.random() == 0.6394267984578837   (baseline)
seed(42) → old generate_random_string() → random.random(): DIFFERENT value  (global state clobbered)
seed(42) → new generate_random_string() → random.random(): 0.6394267984578837  (state preserved)
```

Also verified the new implementation returns length-5 strings drawn from the same pool, and 50 consecutive calls produce 50 distinct values.

## Artifacts (if appropriate):

N/A — behavior change is fully captured by the reproduction above.

PR #7126, opened 2026-07-06
SKY-11836: Avoid reseeding global random in generate_random_string

AronPerez:
## Bug

`generate_random_string()` reseeded Python's module-level `random` PRNG on every call, so unrelated code using `random.*` could have its deterministic or concurrent random stream perturbed.

Linear: https://linear.app/skyvern/issue/SKY-11836/generate-random-string-reseeds-the-global-random-prng-on-every-call

## Root cause

The helper called `random.seed(os.urandom(16))` before drawing with `random.choices(...)`, mutating process-global PRNG state as a side effect.

## Fix

Use a module-local `random.SystemRandom()` instance for the string draw and remove the per-call global reseed. Existing length and character-pool behavior is unchanged.

## Test

Added a regression test that seeds the global PRNG, calls `generate_random_string()`, and asserts `random.getstate()` is unchanged.

Validation:
- `uv sync --python 3.11 --extra server --group dev`
- `uv run pytest tests/unit_tests/test_strings.py -q`
- `uv run ruff check skyvern`
- `uv run ruff format --check skyvern`
- `uv run pre-commit run --files skyvern/utils/strings.py tests/unit_tests/test_strings.py`

### DIFF
--- a/skyvern/utils/strings.py
+++ b/skyvern/utils/strings.py
@@ -1,4 +1,3 @@
-import os
 import random
 import re
 import string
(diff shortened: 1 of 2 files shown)


AronPerez:
Closing as a duplicate. **SKY-11836** already has an open community PR #7076 that predates this one, so we're deferring to the contributor's work there — please review #7076 instead. This PR was opened by an automated agent before we spotted the existing contributor PR; apologies for the duplicate noise. 🙏

PR #7150, opened 2026-07-07
docs(aws): drop stale client parameter from get_file_metadata docstring

anxkhn:
## Description

Removes a stale `client` entry from the `AsyncAWSClient.get_file_metadata`
docstring.

The method signature accepts `uri` and `log_exception`, but the `Args:` block also
documented a `client` parameter that the method does not accept. This keeps the
docstring aligned with the real signature.

## How Has This Been Tested?

- `uvx pre-commit run --files skyvern/forge/sdk/api/aws.py`

## Artifacts

N/A, documentation-only change.


AronPerez (approved):

anxkhn:
Pushed a revision addressing the latest review/CI feedback.

hiSandog:
Removing the stale `client` argument brings the docstring back in line with the actual signature. This might also be a useful place for a lightweight documentation check in CI, such as validating that documented parameter names match inspected function signatures, because this class of drift is easy to reintroduce during refactors. If such checks would be too broad, a focused unit test for this method’s generated API documentation could cover the immediate regression.

anxkhn:
@aronperez thanks for the approval. the branch is rebased onto the latest main. is there anything else needed from me before this can be merged?

PR #7174, opened 2026-07-07
Split block.py 1/8: extract Block base → models/block_base.py (foundation)

AronPerez:
**Part 1/8 — the foundation.** Extracts the `Block` base class and its shared helpers (`capture_block_download_baseline`, `warn_if_file_download_max_steps_low`, `jinja_sandbox_env`, `CURRENT_DATE_FORMAT`, `MAX_STEPS_DOWNLOAD_WARNING_THRESHOLD`) into a new self-contained `models/block_base.py`. `block.py` re-exports them for back-compat. This is the **base of the stack** (targets `main`); every other part imports `Block` from here, which is what breaks the long-standing block.py import cycle.

### The full stack (merge in order, 1 → 8)
1. **1/8** #7174 — foundation: `models/block_base.py` (Block base class + shared helpers), base `main`
2. **2/8** #6979 — branching subsystem → `models/branching.py`
3. **3/8** #7170 — CodeBlock → `models/code_block.py`
4. **4/8** #7171 — parser blocks → `models/parser_blocks.py`
5. **5/8** #6977 — storage blocks → `models/storage_blocks.py`
6. **6/8** #6980 — single blocks (TextPrompt, SendEmail, Wait, TaskV2, HttpRequest, PrintPage, WorkflowTrigger) → `models/misc_blocks.py`
7. **7/8** #6978 — task-block family (BaseTaskBlock + subclasses) → `models/task_blocks.py`
8. **8/8** #7179 — control-flow blocks (ForLoop, WhileLoop, Conditional) → `models/control_flow_blocks.py`

After 8/8, `block.py` is a ~495-line facade: it re-exports everything for zero call-site changes, defines the `BlockSubclasses` discriminated union + dispatch (`get_all_blocks`, `resolve_conditional_merge_edges`), and holds shared helpers only.

Each part's submodule imports `Block` from `block_base` (breaking the historical import cycle); every extracted class stays importable from `models.block` via re-export. Each layer is independently green against the full block-related unit suite.

_(Note: branch slugs still read `..-N7..` from the original 7-part plan; the logical part is the M/8 above.)_


AronPerez:
Addressed the Copilot review in bd5745ec2. Three distinct findings:

**1. `block_base.py:48` — TYPE_CHECKING import of non-existent module (real).** It imported `BaseTaskBlock` from `models.task_blocks`, which is created later in the stack (part 7/8) and doesn't exist against `main`. Repointed to `models.block`, where `BaseTaskBlock` lives in this PR. mypy now resolves it.

**2. `block_base.py:661` — task blocks stopped recording their engine (real regression).** `execute_safe()` replaced the old `isinstance(self, BaseTaskBlock): engine = self.engine` with a polymorphic `self.get_engine()`, and the base returns `None`. The docstring said "Overridden by BaseTaskBlock" but the override was never added — so task blocks silently recorded `engine=None` on their `workflow_run_block`. Added the override on `BaseTaskBlock` plus a regression test (`TestEngineRecordedForTaskBlock`) that fails if it's dropped again.

**3. The ~28 `UnboundLocalError` comments on `with (... as mock_app, patch(..., mock_app), ...)` — false positives.** Python enters each context manager and binds its `as` target *before* evaluating the next context expression (`with A() as a, B(a):` ≡ nested `with`). No error; those tests pass unchanged (61 passing in the two flagged files, 3454 in the full block suite).

PR #7178, opened 2026-07-07
fix: log block failures at warning instead of error

wintonzheng:


AronPerez:
Closing as empty-redundant (precedent #6394). After `git merge origin/main` + taking main's reviewed `files.py`, net diff vs main is empty — its block-failure log-level change is already in main. Nothing left to land.

PR #7180, opened 2026-07-08
fix: log expected custom-select failures at warning, not error

wintonzheng:


AronPerez:
Closing as empty-redundant (precedent #6394). After merge origin/main + taking main's reviewed files.py, net diff vs main is empty — its custom-select log-level change is already in main.

PR #7181, opened 2026-07-08
fix(SKY-12030): bump temporalio lock to 1.30.0 to match pre-freeze prod images

wintonzheng:


AronPerez:
Closing as empty-redundant (precedent #6394). After merge origin/main + taking main's reviewed files.py, net diff vs main is empty — the temporalio 1.30.0 lock bump is already in main.

PR #7182, opened 2026-07-08
fix: log expected download failures at warning instead of error

wintonzheng:


AronPerez:
Closing as empty-redundant (precedent #6394). After merge origin/main + taking main's reviewed files.py, net diff vs main is empty — its download-failure log-level change is already in main.

PR #7184, opened 2026-07-08
fix: classify bad 1Password tokens as 4XX on items endpoint

AronPerez:


AronPerez:
Closing as empty-redundant (precedent #6394). After merge origin/main + taking main's reviewed files.py, net diff vs main is empty — the 1Password 4XX classification (credentials.py + test_workflow_onepassword_errors.py) is already in main.

PR #7185, opened 2026-07-08
fix(frontend): fail safe when broadcasting OAuth credential invalidation

AronPerez:


AronPerez:
Closing as empty-redundant (precedent #6394). After merge origin/main + taking main's reviewed files.py, net diff vs main is empty — the OAuth credential-invalidation fail-safe (useGoogleOAuthCredentials.ts) is already in main.

PR #7190, opened 2026-07-08
fix(SKY-11928,SKY-11929): sandbox checker parity + runtime audit-hook backstop

AronPerez:


AronPerez:
Closing as empty-redundant (precedent #6394). #7190's sandbox-checker feature (code_block_safety.py) is already in main. Its only residual line — latest_recorded_build_test_outcome=None in _advisory_ctx() — is ALSO already present in main's version of this test, so the merge-reconcile produced a duplicate keyword argument (ruff parse error). Net contribution vs main is nil. The reviewed realpath+startswith files.py sanitizer is already in main too. Nothing unique to land.

PR #7193, opened 2026-07-08
fix(frontend): render the UTC hover timestamp in UTC, not local time

anxkhn:
The timestamp tooltips across the app are meant to show the local time as the
visible text and the UTC equivalent on hover. This is the pattern introduced in
#1164 ("make all the timestamp local time and hovering shows UTC timestamp"):
the visible label uses `basicLocalTimeFormat` and the `title=` hover uses
`basicTimeFormat`, which appends a literal `" UTC"` suffix. For example, in
`WorkflowRun.tsx`:

```tsx
<span title={basicTimeFormat(workflowRun.started_at)}>
  Started: {basicLocalTimeFormat(workflowRun.started_at)}
</span>
```

`basicTimeFormat` is used this way as the hover title on ~50 timestamps (run
history, task history, schedules, browser profiles/sessions, workflow runs, TOTP
codes, etc.).

The problem: `basicTimeFormat` did `new Date(time).toLocaleTimeString("en-US")`
with neither `normalizeUtcTimestamp(time)` nor `timeZone: "UTC"`, so it rendered
in the viewer's local time zone while still labeling the output `"UTC"`. Its
sibling `basicLocalTimeFormat` already normalizes the input and pins the resolved
local zone; `basicTimeFormat` was left out when that normalization was added.

Two concrete symptoms:

- A timestamp carrying a `Z`/offset (the API schema examples use
  `2025-01-01T00:00:00Z`) is shown in local time but mislabeled UTC. With
  `TZ=America/New_York`, `2026-03-04T18:30:00Z` renders as
  `Mar 4, 2026 at 1:30:00 PM UTC` when the correct UTC value is `6:30:00 PM`.
- A naive backend timestamp (no `Z`) is parsed by `new Date()` as local, so the
  hover just mirrors the visible local time, making the `"UTC"` tooltip
  redundant and misleading instead of showing the intended UTC instant.

Fix: in `basicTimeFormat`, call `normalizeUtcTimestamp(time)` first, then pass
`timeZone: "UTC"` to both `toLocaleDateString` and `toLocaleTimeString`,
mirroring `basicLocalTimeFormat` but pinned to UTC. No call sites change.

The dead-code sibling `timeFormatWithShortDate` has the same latent issue but has
no callers, so it is intentionally left untouched to keep this change surgical.

### How Has This Been Tested?

Environment: `skyvern-frontend/`, Node 26, `vitest`.

- Extended the existing `basicTimeFormat` unit tests in
  `src/util/timeFormat.test.ts` with two value assertions (the prior tests only
  checked that the string ends with `"UTC"`, which is why the local-vs-UTC bug
  went unnoticed):
  - a `Z`-suffixed input renders the exact UTC value
    (`Wed, Mar 4, 2026 at 6:30:00 PM UTC`), and
  - a naive (no-`Z`) timestamp renders identically to its `Z`-suffixed form.
- Red -> green confirmed: on the pre-fix code with `TZ=America/New_York`, the new
  assertions fail (they get `1:30:00 PM UTC`); after the fix they pass.
- The assertions are timezone-independent: `src/util/timeFormat.test.ts` passes
  under `TZ=UTC`, `TZ=America/New_York` (-5), and `TZ=Asia/Kolkata` (+5:30).
- Full frontend suite green: `npx vitest run` -> 2629/2629 pass, no regressions.
- `npx prettier --check`, `npx eslint --max-warnings 0`, and `npx tsc --noEmit`
  are all clean on the changed files.

### Artifacts (if appropriate):

N/A (behavioral fix to a formatting helper; covered by unit tests above).


AronPerez (approved):

anxkhn:
@aronperez thanks for the approval. the branch is rebased onto the latest main. is there anything else needed from me before this can be merged?

AronPerez:
Let's get this in!

PR #7234, opened 2026-07-09
feat: show login credentials and rotation override on run page

suchintan:


AronPerez:
Closing as redundant (not merged), following the verified #6394 precedent. At PR head 324d69edf and current main 31fbe623c, a normal merge conflicts in six frontend files; taking main for all six produces an exact zero net diff. All 23 PR files are accounted for: 14 are exact current-main blobs, and the nine evolved blobs—including every credential helper/backend symbol and the related tests—were exact main-history blobs before later improvements.

The former `RunWorkflowForm.tsx` survivor must not be preserved: its blob already landed, then #7363 extracted `RotatingCredentialField` and fixed seeded-force → rotation clearing with `null`. Restoring this PR’s older `field.onChange(undefined)` form would reintroduce the React Hook Form no-op regression. The earlier required-check red was only an HTTP 504 while bootstrapping `shellcheck_py` before tests; the single permitted retry (run 29033270412, attempt 2) is now green. This PR does not touch `files.py`, and main retains the realpath + startswith sanitizer. No branch rewrite was pushed. If any intended behavior is missing, this is reopenable.

PR #7278, opened 2026-07-10
[schemas] Keep sanitized parameter keys unique on workflow import

### LINKED ISSUE #7279: Workflow import can produce duplicate parameter keys when sanitized keys collide
### What happened

Importing a workflow whose parameter keys collide after sanitization produces duplicate parameter keys, and any `source_parameter_key` reference to one of those keys ends up pointing at the wrong parameter. It happens silently on the import path (`sanitize_workflow_yaml_with_references` in `skyvern/schemas/workflows.py`, called from `agent_protocol.py`).

### How to reproduce

```python
from skyvern.schemas.workflows import sanitize_workflow_yaml_with_references

wf = {"workflow_definition": {"parameters": [
    {"parameter_type": "workflow", "key": "my-key", "workflow_parameter_type": "string"},
    {"parameter_type": "workflow", "key": "my_key", "workflow_parameter_type": "string"},
    {"parameter_type": "context", "key": "ctx", "source_parameter_key": "my-key"},
], "blocks": []}}

out = sanitize_workflow_yaml_with_references(wf)
print([p["key"] for p in out["workflow_definition"]["parameters"]])
```

Observed: `['my_key_2', 'my_key_2', 'ctx']` (two identical keys), and the context parameter's `source_parameter_key` comes out as `my_key_2`.

Expected: `['my_key', 'my_key_2', 'ctx']`, and `source_parameter_key` should be `my_key`.

### Why

`my-key` sanitizes to `my_key`, which already exists, so the second key is suffixed to `my_key_2`. The mapping becomes `{'my-key': 'my_key', 'my_key': 'my_key_2'}`. Step 4 then applies the mapping once per entry and recurses into every string, so it rewrites each parameter's own already-final `key` field and chains a later substitution onto an earlier one. `error_code_mapping` was already protected against this exact chaining with `_rewrite_error_code_mapping_refs_atomic`, but the parameter references were not.

### Fix

Opened #7278. It resolves the direct string references against the whole mapping in one pass (first match wins, so nothing chains) and leaves the parameter `key` field alone since it is already final. Regression tests cover the collision, the reference chaining, and the cases that should stay unchanged.

@suchintan @wintonzheng if this looks right to you I am happy to adjust anything in the PR.

abhay-codes07:
## What is wrong

`sanitize_workflow_yaml_with_references` in `skyvern/schemas/workflows.py` runs on the public workflow-import path (`agent_protocol.py`), so it touches every workflow YAML/JSON that gets posted. Step 4 rewrites parameter-key references by looping over the key mapping and recursing into every string in the parameters list, once per mapping entry. That has two side effects I did not expect:

1. It rewrites each parameter's own `key` field, even though `_sanitize_parameters` already finalized it in Step 2.
2. The substitutions run one after another, so a reference can chain from one mapping entry into the next.

When two distinct, valid keys sanitize into a collision this goes wrong. Take `my-key` and `my_key`. `my-key` normalizes to `my_key`, which already exists, so it gets suffixed to `my_key_2`. The mapping ends up `{'my-key': 'my_key', 'my_key': 'my_key_2'}`. Then:

- The first parameter's key was set to `my_key` in Step 2. The second mapping entry (`my_key -> my_key_2`) matches it and rewrites it to `my_key_2`, so now both parameters have the key `my_key_2`.
- A `source_parameter_key: "my-key"` gets rewritten `my-key -> my_key` by the first entry, then `my_key -> my_key_2` by the second, so it points at the wrong parameter.

All of this is silent. A workflow that was fine on the way in comes out with duplicate keys or a mis-wired reference.

## Reproduction

```python
from skyvern.schemas.workflows import sanitize_workflow_yaml_with_references

wf = {"workflow_definition": {"parameters": [
    {"parameter_type": "workflow", "key": "my-key", "workflow_parameter_type": "string"},
    {"parameter_type": "workflow", "key": "my_key", "workflow_parameter_type": "string"},
    {"parameter_type": "context", "key": "ctx", "source_parameter_key": "my-key"},
], "blocks": []}}

out = sanitize_workflow_yaml_with_references(wf)
print([p["key"] for p in out["workflow_definition"]["parameters"]])
# before: ['my_key_2', 'my_key_2', 'ctx']   <- duplicate
# after:  ['my_key', 'my_key_2', 'ctx']
```

The `source_parameter_key` on the context parameter comes out as `my_key_2` before the fix and `my_key` after it.

## The fix

The interesting thing is the codebase already knows about this hazard. `_rewrite_error_code_mapping_refs_atomic` was added specifically so error_code_mapping substitutions do not chain. Step 4's parameter references were just never given the same treatment.

I resolve the direct string references (`source_parameter_key` and friends) against the whole mapping in a single pass that stops at the first match, so nothing chains, and I leave each parameter's own `key` field alone since it is already final after Step 2. The per-entry `_replace_direct_string_in_value` call in Step 4 is replaced by one `_replace_direct_string_param_refs` pass after the loop.

## Tests

`tests/unit/test_workflow_parameter_key_sanitization.py` covers:

- colliding sanitized keys stay unique (no duplicate)
- `source_parameter_key` follows the renamed key without chaining
- a reference to the suffixed parameter still resolves correctly
- a plain single rename still updates its reference
- already-valid keys and references are untouched

The two collision tests fail on the current code and pass with the fix. I also ran the existing `test_workflow_error_code_mapping_inheritance.py`, `test_workflow_parameter_validation.py`, and `test_workflow_schema_field_preservation.py` suites and they stay green. `ruff check` and `ruff format` are clean.

Closes #7279

### DIFF
--- a/skyvern/schemas/workflows.py
+++ b/skyvern/schemas/workflows.py
@@ -119,6 +119,28 @@ def _replace_direct_string_in_value(value: Any, old_key: str, new_key: str) -> A
     return value
 
 
+def _replace_direct_string_param_refs(value: Any, key_mapping: dict[str, str], *, _in_key_field: bool = False) -> Any:
+    """Rewrites exact-match direct string references to parameter keys (e.g. source_parameter_key).
+
+    Resolves each string against the whole mapping in a single pass and stops at the first match,
+    so a collision-driven rename chain (my-key -> my_key AND my_key -> my_key_2) can't compound a
+    reference into the wrong target. The parameter's own `key` field is never rewritten here: it is
+    already finalized by _sanitize_parameters, and re-applying the mapping to it would collapse two
+    distinct keys into the same name.
+    """
+    if isinstance(value, str):
+        if _in_key_field:
+            return value
+        return key_mapping.get(value, value)
+    elif isinstance(value, dict):
+        return {
+            k: _replace_direct_string_param_refs(v, key_mapping, _in_key_field=(k == "key")) for k, v in value.items()
+        }
+    elif isinstance(value, list):
+        return [_replace_direct_string_param_refs(item, key_mapping) for item in value]
+    return value
+
+
 def _make_unique(candidate: str, seen: set[str]) -> str:
     """Appends a numeric suffix to make candidate unique within the seen set.
 
@@ -350,15 +372,21 @@ def sanitize_workflow_yaml_with_references(workflow_yaml: dict[str, Any]) -> dic
             workflow_definition["parameters"] = _replace_references_in_value(
                 workflow_definition["parameters"], old_key, new_key
             )
-            # Also update direct string references (e.g., source_parameter_key)
-            workflow_definition["parameters"] = _replace_direct_string_in_value(
-                workflow_definition["parameters"], old_key, new_key
-            )
 
         if isinstance(workflow_definition.get("workflow_system_prompt"), str):
             workflow_definition["workflow_system_prompt"] = _replace_references_in_value(
                 workflow_definition["workflow_system_prompt"], old_key, new_key
             )
+
+    # Update direct string references to parameter keys (e.g. source_parameter_key) in one atomic
+    # pass. Doing this per-mapping-entry sequentially chained a collision rename (my-key -> my_key
+    # then my_key -> my_key_2) onto the already-renamed reference, and it also rewrote each param's
+    # own final `key` field back into a duplicate. _replace_direct_string_param_refs leaves the key
+    # field alone and resolves each reference against the whole mapping exactly once.
+    if param_key_mapping and "parameters" in workflow_definition:
+        workflow_definition["parameters"] = _replace_direct_string_param_refs(
+            workflow_definition["parameters"], param_key_mapping
+        )
 
     # Rewrite workflow-level error_code_mapping atomically so substitutions don't chain
     # (e.g. foo-bar -> foo_bar and foo_bar -> foo_bar_2 must not combine into foo_bar_2).
--- a/skyvern/schemas/workflows.py
+++ b/skyvern/schemas/workflows.py
@@ -74,15 +74,50 @@ def _get_text_prompt_model_name_by_llm_key() -> dict[str, str]:
     return reverse_mapping
 
 
-def _replace_references_in_value(value: Any, old_key: str, new_key: str) -> Any:
-    """Recursively replaces Jinja references in a value (string, dict, or list)."""
-    if isinstance(value, str):
-        return replace_jinja_reference(value, old_key, new_key)
-    elif isinstance(value, dict):
-        return {k: _replace_references_in_value(v, old_key, new_key) for k, v in value.items()}
-    elif isinstance(value, list):
-        return [_replace_references_in_value(item, old_key, new_key) for item in value]
-    return value
+def _make_sentinels(substitutions: list[tuple[str, str]]) -> list[tuple[str, str, str]]:
+    """Pairs each (old, new) substitution with a unique, un-typeable sentinel token.
+
+    The sentinels let a whole substitution set be applied atomically: every old identifier is
+    swapped to its sentinel first, then every sentinel to its new identifier. That breaks the
+    chain when one substitution's new identifier matches another's old one (e.g. foo-bar ->
+    foo_bar AND foo_bar -> foo_bar_2), which a naive sequential rewrite would compound.
+    """
+    return [(old, new, f"\x00SKY_SANITIZE_{i}\x00") for i, (old, new) in enumerate(substitutions)]
+
+
+def _apply_atomic_jinja_substitutions(text: str, sentinels: list[tuple[str, str, str]]) -> str:
+    """Applies a sentinel-guarded set of Jinja reference substitutions to a single string."""
+    for old, _new, sentinel in sentinels:
+        text = replace_jinja_reference(text, old, sentinel)
+    for _old, new, sentinel in sentinels:
+        text = text.replace(sentinel, new)
+    return text
+
+
+def _rewrite_jinja_refs_atomic(value: Any, substitutions: list[tuple[str, str]]) -> Any:
+    """Atomically rewrites Jinja references in a value (string, dict, or list).
+
+    Every reference is resolved against the whole substitution set in one pass via unique
+    sentinels, so a collision rename chain (my-key -> my_key AND my_key -> my_key_2) can't
+    compound a reference into the wrong target. Order still matters within the set: place the
+    more specific substitution (e.g. {label}_output) before the bare label so the bare pass
+    does not partially match an already-sentinelled reference.
+    """
+    if not substitutions:
+        return value
+
+    sentinels = _make_sentinels(substitutions)
+
+    def _walk(v: Any) -> Any:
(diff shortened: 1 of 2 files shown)


skyvern/schemas/workflows.py
```diff
@@ -350,16 +372,22 @@ def sanitize_workflow_yaml_with_references(workflow_yaml: dict[str, Any]) -> dic
...
-            workflow_definition["parameters"] = _replace_direct_string_in_value(
-                workflow_definition["parameters"], old_key, new_key
-            )
 
         if isinstance(workflow_definition.get("workflow_system_prompt"), str):
             workflow_definition["workflow_system_prompt"] = _replace_references_in_value(
                 workflow_definition["workflow_system_prompt"], old_key, new_key
             )
 
+    # Update direct string references to parameter keys (e.g. source_parameter_key) in one atomic
```
AronPerez: This only makes exact-string fields atomic, while the preceding per-entry Jinja pass still rewrites `{{ my-key }}` to `{{ my_key }}` and then `{{ my_key_2 }}`, so colliding-key references in defaults, blocks, and the system prompt remain miswired.

tests/unit/test_workflow_parameter_key_sanitization.py
```diff
@@ -0,0 +1,97 @@
...
+                {"key": "ctx", "parameter_type": "context", "source_parameter_key": "my-key"},
+            ],
+            "blocks": [],
+        }
+    }
+    sanitized = sanitize_workflow_yaml_with_references(workflow_yaml)
+    params = sanitized["workflow_definition"]["parameters"]
+    ctx = next(param for param in params if param["parameter_type"] == "context")
+    # "my-key" was renamed to "my_key"; the reference must land there, not chain to "my_key_2".
+    assert ctx["source_parameter_key"] == "my_key"
```
AronPerez, line 46: Please add a collision regression with a Jinja reference such as `default_value: "{{ my-key }}"` and ideally a block reference so the remaining chained path is exposed.

AronPerez (changes requested):

skyvern/schemas/workflows.py
abhay-codes07: Good catch, you are right. The atomic pass I added only covered the exact-string source_parameter_key field, while the per-entry Jinja loop above it still chained. I reworked both the label pass (Step 3) and the parameter-key pass (Step 4) to resolve every reference against the whole substitution set in a single sentinel-guarded pass, so a collision rename can no longer compound {{ my-key }} into {{ my_key_2 }} in defaults, blocks, or the workflow system prompt. This is the same sentinel approach already used for error_code_mapping, now shared through a small helper. The direct-string output-key and source_parameter_key rewrites are folded into one first-match resolution that leaves each parameter's own key field alone.

tests/unit/test_workflow_parameter_key_sanitization.py
abhay-codes07: Added both: test_jinja_default_value_reference_follows_renamed_key_without_chaining covers a default_value: "hello {{ my-key }}" reference, and test_jinja_block_reference_follows_renamed_key_without_chaining covers a block navigation_goal referencing both {{ my-key }} and {{ my_key }} in the same string (asserting they land on my_key and my_key_2 respectively). Both fail on the old per-entry code and pass with the atomic pass.

abhay-codes07:
Pushed 673ae5b. Both Jinja passes (block labels and parameter keys) are now atomic, so the remaining chained path you flagged is fixed for defaults, blocks, and the system prompt, not just source_parameter_key. Also added the two collision regressions with Jinja references. Full sanitizer suite is green locally (80 passed). Thanks for the review.

skyvern/schemas/workflows.py
```diff
@@ -74,15 +74,50 @@ def _get_text_prompt_model_name_by_llm_key() -> dict[str, str]:
...
+    swapped to its sentinel first, then every sentinel to its new identifier. That breaks the
+    chain when one substitution's new identifier matches another's old one (e.g. foo-bar ->
+    foo_bar AND foo_bar -> foo_bar_2), which a naive sequential rewrite would compound.
+    """
+    return [(old, new, f"\x00SKY_SANITIZE_{i}\x00") for i, (old, new) in enumerate(substitutions)]
+
+
+def _apply_atomic_jinja_substitutions(text: str, sentinels: list[tuple[str, str, str]]) -> str:
+    """Applies a sentinel-guarded set of Jinja reference substitutions to a single string."""
+    for old, _new, sentinel in sentinels:
```
AronPerez, line 90: Cycling every substitution's pattern per string defeats re's 512-entry pattern cache: at 520 renames I measured 21.5s here versus 0.78s on main, and 54.6s versus 1.8s at 800. This runs synchronously inside the async import handlers on an uncapped request body, so a single authenticated POST stalls the worker; compiling each pattern once per substitution set restores the old cost.

skyvern/schemas/workflows.py
```diff
@@ -298,67 +340,47 @@ def sanitize_workflow_yaml_with_references(workflow_yaml: dict[str, Any]) -> dic
...
+        for jinja_field in ("blocks", "parameters", "workflow_system_prompt"):
+            if jinja_field in workflow_definition:
+                workflow_definition[jinja_field] = _rewrite_jinja_refs_atomic(
+                    workflow_definition[jinja_field], label_substitutions
+                )
+
+    # Step 4: Update all parameter key references. Same anti-chaining reasoning as Step 3: a collision
+    # rename (my-key -> my_key AND my_key -> my_key_2) must not compound {{ my-key }} into
+    # {{ my_key_2 }} across defaults, blocks, and the system prompt, so resolve them in one atomic pass.
+    param_substitutions = list(param_key_mapping.items())
```
AronPerez, line 367: Label and parameter substitutions are still applied as two sequential passes, so a bare reference whose sanitized label equals a suffixed parameter's old key resolves to the parameter instead of the block. Pre-existing rather than introduced here, but the error_code_mapping rewrite below already builds one combined set, and doing the same here would satisfy the invariant this comment states.

tests/unit/test_workflow_parameter_key_sanitization.py
```diff
@@ -0,0 +1,145 @@
...
+            "blocks": [],
+        }
+    }
+    sanitized = sanitize_workflow_yaml_with_references(workflow_yaml)
+    params = sanitized["workflow_definition"]["parameters"]
+    greeting = next(param for param in params if param["key"] == "greeting")
+    assert greeting["default_value"] == "hello {{ my_key }}"
+
+
+def test_jinja_block_reference_follows_renamed_key_without_chaining() -> None:
```
AronPerez, line 108: nit (optional): the label pass became atomic too, and that is the silent half - a label collision yields a valid reference to the wrong block instead of a 422 - but no test covers a label collision chain.

AronPerez (changes requested):
The chaining fix verifies correct, but the rewrite introduces a perf regression on the import path.

- Original change-request is discharged; the chaining fix verifies clean.
- New: past 512 renames the import path is ~27x slower than main.
- That call is sync inside async handlers, so it stalls the worker.
- Branch conflicts with main and needs a rebase before merge.
- Follow-up (pre-existing): loop_over_parameter_key is never remapped.

PR #7322, opened 2026-07-12
fix(SKY-12124): unwrap raw_arguments wrapper at the FastMCP tool-call boundary

AronPerez:
## Description

The Workflow Copilot LLM and remote MCP clients sometimes serialize an entire tool-call payload as `{"raw_arguments": {...actual args...}}` instead of spreading the arguments. Because no `skyvern_` tool declares a `raw_arguments` parameter, this fails **pydantic signature validation at the FastMCP tool-call boundary**:

```
ValidationError: 2 validation errors for call[skyvern_navigate]
url            Missing required argument
raw_arguments  Unexpected keyword argument
```

It is one serialization bug that surfaces across every tool the wrapper reaches (`skyvern_browser_session_create`, `skyvern_run_task`, `skyvern_navigate`, …).

**Fix:** add `ArgRepairMiddleware`, a FastMCP `Middleware.on_call_tool` on the shared `mcp` app, that conservatively unwraps a `raw_arguments` object (or a JSON-object string) into top-level args **before** validation runs.

Placed on the shared `mcp` app so it covers **both** client surfaces in one fix:
- the in-memory Workflow Copilot overlay client (`forge/sdk/copilot/mcp_adapter.py`), and
- remote/HTTP MCP clients (`skyvern-api`).

Conservative guardrails (never mask a genuinely malformed call):
- Only a dict / JSON-object `raw_arguments` is unwrapped; a non-object value is left untouched so the call still errors.
- `raw_arguments` is unwrapped only when it is the sole key. Any sibling makes the call ambiguous, so the payload is left untouched for normal validation to reject.
- The whole repair pass is wrapped so a repair bug degrades to passthrough, never a failed tool call.

⚠️ Touches the **MCP tool-call boundary**.

Fixes SKY-12124, SKY-12125, SKY-12127.

## How Has This Been Tested?

New `tests/unit/test_mcp_arg_repair.py` drives the real `mcp` app through an in-memory FastMCP `Client`, so the full middleware + pydantic-validation path is exercised (not the tool functions in isolation):

- a dict wrapper and a JSON-string wrapper are unwrapped and the call succeeds;
- a wrapper with any sibling argument is left untouched and still rejected;
- a non-object `raw_arguments` still errors (proves malformed calls are not masked).

`uv run pytest tests/unit/test_mcp_arg_repair.py -v` passes the 5 focused base tests; all-files Ruff check/format is clean.

n/a


AronPerez:
Rebased onto current `main` and revalidated the shared FastMCP boundary. CI is green.

The conflict with the newer structured argument validator was resolved by preserving middleware order: telemetry -> arg repair -> argument validation. Raw unwrap remains sole-key-only, so ambiguous or malformed calls still reach validation and return `INVALID_INPUT`.

@wintonzheng @andrewneilson could one of you review/approve this base PR first? #7324 remains stacked on it.

AronPerez:
Final review head is `0b7bd192b9babde1666e7b7bf7e4c558d7401eeb`, rebased directly onto current `main` (`c9aabd38014fcf32d3e749944e33d5e13907242b`). All GitHub checks are green; focused boundary suite is 5/5 and all-files Ruff check/format is clean.

The repair is on the shared FastMCP `mcp` app as `Middleware.on_call_tool`, before structured argument validation. `raw_arguments` unwrap is sole-key-only; sibling/invalid/unknown arguments remain untouched and still error.

@wintonzheng @andrewneilson — this base PR now only needs one independent approval; squash auto-merge is armed.

skyvern/cli/mcp_tools/arg_repair.py
```diff
@@ -0,0 +1,229 @@
...
+    SOLE argument and holds a dict (or a JSON-object string) it is unambiguously
+    the misplaced payload — lift its keys up.
+
+    Restricted to the sole-key case on purpose: if real sibling args sit next to
+    ``raw_arguments`` the call is ambiguous (the blob may be garbage, or duplicate
+    a top-level arg), so it is left to error rather than merging stray keys into
+    an otherwise-valid call. A non-object (or over-long) ``raw_arguments`` is
+    likewise left to error instead of being silently swallowed.
+    """
+    if list(arguments) != ["raw_arguments"]:
```
andrewneilson, line 73: The Copilot adapter injects `session_id` before this middleware, so browser calls no longer satisfy the sole-key guard and the advertised `raw_arguments` fix does not apply. Normalize at the overlay entry before pre-hooks and forced arguments, while retaining this middleware for direct clients.

andrewneilson (changes requested):
Requesting changes: Copilot overlay normalization occurs after its policy hooks and injected arguments.

- Browser tools retain the original validation failure.
- Non-browser hooks inspect a different payload than FastMCP dispatches.

PR #7324, opened 2026-07-12
fix(SKY-12048): coerce unambiguous MCP tool-call arg-shape mistakes at the boundary

AronPerez:
## Description

> **Stacked on #7322** (`ao/skyvern-75/sky-12124-mcp-raw-arg-unwrap`). Review/merge that first; this PR targets that branch and its diff is only the coercion registry.

Extends the FastMCP `ArgRepairMiddleware` (introduced in the base PR) with conservative, **per-tool** coercions for arg-shape mistakes the copilot LLM and remote MCP clients make. Each case has exactly one unambiguous reading, so repairing it at the boundary cannot mask intent:

| Tool | Bad payload (real, from Datadog) | Repair | Tickets |
|------|----------------------------------|--------|---------|
| `skyvern_code_block_lint` | `parameter_keys='["a","b"]'` (stringified list) or `parameter_keys="a"` (bare key) | → `list[str]` | SKY-12048, SKY-12049 |
| `skyvern_extract` | `schema={"type":"object",…}` (JSON object) | → its JSON string | SKY-12338 |
| `skyvern_run_task` | `data_extraction_schema={"type":"object",…}` (JSON object) | → its JSON string | SKY-12789 |
| `skyvern_browser_profile_get` | `profile_id="bp_…"` | → `browser_profile_id` | SKY-12581 |
| `skyvern_workflow_status` | `workflow_run_id="wr_…"` | → `run_id` | linked to SKY-12048 |
| `skyvern_block_validate` | a block definition misnamed `block` / `definition` / `block_definition` / `block_yaml` | → `block_json` | SKY-11133 |

**Why `block_validate` again (SKY-11133):** the earlier fix (#6658) added this alias only in the copilot pre-hook, which never runs for **remote/HTTP** MCP clients — so the error reopened. Moving the alias to the shared `mcp`-app boundary covers the remote path too, which is where it kept recurring.

**Deliberately NOT coerced (kept erroring):** a full block *definition* sent to `skyvern_block_schema` (which takes a `block_type` string, not a definition). The real payloads show the model passing a complete definition as `block_json` / `definition` with no `block_type` — it meant `skyvern_block_validate`. Coercing that would mask a **wrong-tool** selection bug (exactly how SKY-11133 first regressed). SKY-12140 / SKY-12141 are therefore routed to the prompt/tool-description track, not fixed here, and a regression test asserts these calls still error.

⚠️ Touches the **MCP tool-call boundary**. Coercions are conservative: non-string / non-object values, and anything without a single unambiguous reading, pass through untouched.

Fixes SKY-12048, SKY-12049, SKY-12338, SKY-11133.

## How Has This Been Tested?

Added to `tests/unit/test_mcp_arg_repair.py`:

- unit tests of `repair_tool_arguments` for each coercion (JSON-list string, python-repr list string, bare string → list; dict → JSON string; alias → `block_json`; canonical+alias conflicts and ambiguous aliases remain rejected);
- passthrough/no-mask tests (real `parameter_keys` list untouched, non-string scalar untouched, `block_schema` definition arg untouched);
- boundary integration through an in-memory FastMCP `Client` on the real `mcp` app: `block_validate(block=…)` validates after promotion, `code_block_lint(parameter_keys='["value"]')` validates, and `block_schema(block_json=…)` still errors.

`uv run --extra server pytest tests/unit/test_mcp_arg_repair.py` and the MCP block/lint/registry suites pass (47 passed); pre-commit (ruff, mypy, isort) clean.

## Follow-ups

- **SKY-12348** — widen the `skyvern_extract` `schema` param to `str | dict | None` and drop the middleware dict→string coercion (the cleaner layering; `parse_extract_schema` already accepts a dict). Surfaced by the high-effort review gate; non-blocking since the dict-only coercion here already fixes the correctness issue.

## Review-gate hardening (post-review commit)

A high-effort masking review flagged three ways the coercions could mask a legitimately-erroring call; all fixed with no-mask regression tests:
- `raw_arguments` is now unwrapped **only when it is the sole argument** (a sibling arg → left to error, no stray-key injection);
- `parameter_keys` coercion accepts **only an all-strings list** (nested / non-string / unparseable → left to error, no phantom keys);
- `schema` coercion is **dict-only** (a JSON array is not a valid schema object → left to error);
- added a length cap so unbounded remote input is never parsed before validation;
- a follow-up codex masking pass surfaced a 4th same-class edge: a JSON-**object**-looking `parameter_keys` string (e.g. `'{"a":1}'`) is now left to error instead of being wrapped into a phantom key (no real key starts with `{`).
- a final codex round closed two consistency gaps: (a) the DoS length-cap now also bounds the `raw_arguments` parse and the `json.dumps` outputs (schema / block dict), so an unbounded string never crosses the boundary; (b) `block_json` alias promotion no longer silently resolves ambiguity — a distinct alias alongside a present `block_json`, two distinct payloads across aliases, a malformed non-string `block_json`, or a non-promotable alias value are all left to error instead of one being discarded/overwritten. This is intentionally stricter than the copilot pre-hook copy (which runs first on the copilot path); this middleware is the remote-facing boundary.

### Copilot-path safety (no new errors on the copilot path)

The stricter `block_json` behavior affects **only remote/HTTP clients**. On the copilot path, `validate_block`'s pre-hook (`forge/sdk/copilot/tools/mcp_hooks.py::_normalize_block_json_alias`) runs **before** this middleware and unconditionally strips **every** block alias — so by the time this middleware runs there are no aliases left for it to reject, making it a **no-op** on the copilot path. Verified for every ambiguity case the middleware now errors on (conflicting-distinct, distinct-alongside-canonical, non-string-canonical, non-promotable, identical-across-names): the copilot call is unchanged and still passes exactly as before. The boundary registry remains tool-specific; unknown parameters and conflicting canonical/alias pairs are never discarded or overwritten. This coverage depends on the two `_BLOCK_JSON_ALIASES` tuples staying identical — asserted by `test_copilot_prehook_covers_every_middleware_block_alias_error` and flagged in the sync comment so a future resync can't silently re-weaken the remote boundary.

n/a


AronPerez:
Rebased onto the updated #7322 without changing the inherited production patch. The only post-rebase test adjustments align no-mask assertions with `main`'s structured `INVALID_INPUT` validator and make the acceptance case exercise a Python-list repr string through the real shared FastMCP app.

Acceptance payload:
```json
// before
{"code":"value = 1\n","parameter_keys":"['value']"}
// after boundary repair
{"code":"value = 1\n","parameter_keys":["value"]}
```
The real `skyvern_code_block_lint` boundary call completes with `is_error=false`; focused suite: 31 passed. Wrong-tool `skyvern_block_schema(block_json=...)`, ambiguous `raw_arguments`, unknown aliases, and malformed shapes remain rejected.

@wintonzheng @andrewneilson please review this after the base PR is approved/merged.

AronPerez:
Datadog grooming (2026-07-21) — three more real payloads from the last 5 days that belong in this coercion registry:

- `skyvern_run_task`: `data_extraction_schema={'type':'object',...}` (JSON object where the contract wants a JSON string) — SKY-12789. Occurred after #13652 merged, since key-checking doesn't catch value-shape. Mirror of the existing `skyvern_extract.schema` coercion.
- `skyvern_browser_profile_get`: `profile_id=...` instead of `browser_profile_id` — SKY-12581. Unambiguous rename, same class as the `block_json` misnames.
- `skyvern_workflow_status`: wrong-key ValidationError seen in Datadog error tracking (no Linear ticket in window).

All three linked to SKY-12048 in Linear; no duplicate worker spawned — extending this PR is the single-root-cause fix.

AronPerez:
Follow-up on the `skyvern_workflow_status` case, per manager review: verified the raw payload — `workflow_run_id='wr_514972324332793994'` passed where the contract wants `run_id`. The value is a workflow-run id and the tool is workflow_status, so this is a genuine alias (same entity, more explicit name), in-scope for a conservative misname→canonical repair. Wrong-tool / hallucinated params should still raise — this one isn't that.

AronPerez:
Addressed the Datadog follow-ups on head `6766c1d86`:

- `skyvern_run_task.data_extraction_schema`: dict -> bounded JSON string (SKY-12789)
- `skyvern_browser_profile_get.profile_id`: exact alias -> `browser_profile_id` (SKY-12581)
- `skyvern_workflow_status.workflow_run_id`: exact verified alias -> `run_id`

All three are tool-specific registry entries in the shared FastMCP middleware. Boundary tests exercise each repaired call before pydantic validation. Canonical+alias conflicts, unknown siblings/keys, list-shaped schemas, oversized dicts, and wrong-tool `skyvern_block_schema(block_json=...)` remain unmodified and still reject.

Focused suite: `47 passed`. All-files pre-commit is green. The full unit run reached `16,398 passed` with only the unrelated timing-flaky `test_timed_out_deletion_stops_before_cleanup_returns`; that test passes in isolation.

AronPerez:
Final stacked head is `fb788e8bf3e54a77bbd63c741275ecf03ca3c216`, based exactly on #7322 head `0b7bd192b9babde1666e7b7bf7e4c558d7401eeb`. All GitHub checks are green; focused boundary suite is 47/47 and all-files Ruff check/format is clean.

SKY-12048 acceptance evidence through the real shared FastMCP app:
```json
// before
{"code":"value = 1\n","parameter_keys":"['value']"}
// after boundary repair
{"code":"value = 1\n","parameter_keys":["value"]}
```
`skyvern_code_block_lint` completes with `is_error=false`. Unknown parameters, canonical/alias conflicts, malformed shapes, and wrong-tool `skyvern_block_schema` payloads still reject.

@wintonzheng @andrewneilson — please review after #7322 auto-merges.

PR #7345, opened 2026-07-13
fix(SKY-11961): catch typed captcha-solve failures before the generic handler arm

AronPerez:


AronPerez:
Closing as **redundant** (sync-mgr #6394 redundancy pass). Verified this sync-mirror's content is already on `main`: reconciling with main (taking main on all overlaps) yields net-vs-main = 0 (main is a superset), and every symbol this branch introduces — e.g. `CaptchaNotSolvedInTime (exceptions.py)` — already exists on main. It landed via squash/re-mirror commits (different patch-id, identical content), which is why a patch-id check alone shows 'unique'. No unique contribution remains to merge. Please reopen if this analysis is mistaken.

PR #7362, opened 2026-07-13
refactor(SKY-11697): extract _generate_step_actions from ForgeAgent.agent_step

AronPerez:


AronPerez:
Closing as redundant (not merged), following the verified #6394 precedent. At PR head 3adb1be31 and main 94d5156f7, merging main and taking main for the sole conflicting file (`skyvern/forge/agent.py`) produces an exact zero net diff. The PR-added `_generate_step_actions` helper and its call/three-value wiring are byte-for-byte present on main; the existing 12-case agent-step characterization suite passes there. This PR does not touch `files.py`, and the main sanitizer remains intact. No branch rewrite was pushed. If any intended behavior is missing, this is reopenable.

PR #7367, opened 2026-07-13
Prefer the real form submit over a consent/opt-out overlay when a click is intercepted (SKY-12352)

AronPerez:


AronPerez:
Closing as redundant after cumulative fix PR #7371 merged (precedent: #6394). Verified in a clean detached merge of current main: net diff is empty when taking main's fixed tests/unit/test_pdf_fill_block.py, and every symbol/content addition from this PR is present on main. The shared vulnerable blob 8310183b was superseded only by the safe one-line regex in main blob 2d9bbd7b; CodeQL alert #329 is absent on main. Reopenable if any unique delta is found.

PR #7368, opened 2026-07-13
refactor(SKY-11698): extract _execute_step_actions + _BackgroundArtifactTaskTracker from ForgeAgent.agent_step

AronPerez:


AronPerez:
Closing as redundant after cumulative fix PR #7371 merged (precedent: #6394). Verified in a clean detached merge of current main: net diff is empty when taking main's fixed tests/unit/test_pdf_fill_block.py, and every symbol/content addition from this PR is present on main. The shared vulnerable blob 8310183b was superseded only by the safe one-line regex in main blob 2d9bbd7b; CodeQL alert #329 is absent on main. Reopenable if any unique delta is found.

PR #7417, opened 2026-07-14
Light theme for the workflows code editors

celalzamanoglu:


AronPerez:
Redundancy sweep complete after #7446 merged. The original PR diff contained only tests/unit/forge/sdk/db/test_self_heal_schema.py, changing the migration assertions to private revision 3b9d7a4c1e2f; CI failed on that exact mismatch because the checkout head is 2b8e37d98d97. After rebasing onto origin/main and resolving the main-owned files (including skyvern/forge/sdk/api/files.py and this schema test), HEAD equals origin/main at 75b4a5486 and the net diff is empty. Main already contains the shared ScriptDirectory/heads/revision/down_revision symbols, durable docstring lookup, and public 2b8e37d98d97 ancestor check (git grep origin/main). Focused validation: 21 schema tests passed; Alembic reports one head, 2b8e37d98d97. Closing as redundant per precedent #6394.

PR #7421, opened 2026-07-14
SKY-12451: add workflow-run selection and tag management

AronPerez:

PR #7426, opened 2026-07-14
fix(SKY-12441): default studio layout mirrors legacy edit/run views

celalzamanoglu:


AronPerez:
Closing as redundant/stale (precedent #6394 + #7421 take-main class). Provenance: main's 'flex min-w-0 items-center gap-2' in WorkflowPage.tsx/RunHistory.tsx was introduced by #7448 (merged 17:45Z), which is newer than this PR's mirror base. This PR's entire remaining net diff reverts that line to the older 'flex-col gap-1' form (and drops min-w-0 in WorkflowPage). Its own SKY-12441 studio-layout content is already on main via the cumulative sync merges. Nothing genuine to land.

PR #7429, opened 2026-07-14
Copilot: stop treating a login prefix as goal-complete when the goal is a downstream page (SKY-12449)

andrewneilson:


AronPerez:
Closing as redundant, following precedent #6394. After merging origin/main at 75b4a5486, taking main's files.py and self-heal test, and normalizing the two conflict-only layout hunks to current origin/main, the final tree has a zero-file net diff (both worktree and index comparisons to origin/main exit 0). Spot-checks confirm the original additions are already on origin/main: RunBulkActionBar, RunRowContextMenu, RunTagsEditor, useRunHealEpisodesQuery, BlockHealPanel/RunHealChip/healStatus, and the run-tag tests. The known CI failure is resolved on main: 3b9d7a4c1e2f is absent and the durable docstring-lookup test is present. No push needed for the redundant PR.

PR #7432, opened 2026-07-15
fix: harden the shared secret detection/redaction pattern

pedrohsdb:


AronPerez:
Closing as redundant per precedent #6394. Evidence: after rebasing the PR head onto origin/main (75b4a5486, #7446), the final pushed tree at 1c8a654d1 has an empty net diff versus origin/main; skyvern/forge/sdk/api/files.py and tests/unit/forge/sdk/db/test_self_heal_schema.py are byte-identical to main. An audit of 1,700 identifier candidates from the original PR diff found no symbol absent from origin/main, including the secret-redaction, run-tag, and self-heal symbols. The only failed CI test was the known private migration-ID assertion (3b9d7a4c1e2f vs main's 2b8e37d98d97), resolved by taking main's schema test. Focused validation passed: 2 frontend files / 18 tests, and the self-heal schema file / 21 tests.

PR #7555, opened 2026-07-19
[schemas] Fix chained Jinja reference rewrites when sanitized workflow identifiers collide

### LINKED ISSUE #7554: Colliding sanitized workflow identifiers chain Jinja reference rewrites onto the wrong parameter
Found while reading `sanitize_workflow_yaml_with_references` after Skyvern-AI/skyvern#7279. This is a different defect from [#7278](<https://github.com/Skyvern-AI/skyvern/issues/7278>) (which covers the direct-string pass) — the Jinja reference rewrites have their own chaining problem.

The rewrites in steps 3/4 run once per renamed identifier, sequentially. So when a collision-driven rename overlaps an earlier one, the later substitution re-hits the earlier one's output. Import this:

```yaml
workflow_definition:
  parameters:
    - key: user email    # sanitized to user_email
    - key: user_email    # valid, but now colliding -> suffixed to user_email_2
  blocks:
    - block_type: task
      label: send
      navigation_goal: "Send to {{ user email }} and CC {{ user_email }}"
```

The goal comes out as `Send to {{ user_email_2 }} and CC {{ user_email_2 }}` — both references now point at the second parameter, and the workflow runs with the wrong value. No error at import or run time (default lax templating). Block labels hit the same thing through the `{label}_output` rewrites: two blocks whose labels collide end up with their output references collapsed onto one block, e.g. `{{ foo/bar_output }} vs {{ foo_bar_output }}` becomes `{{ foo_bar_2_output }} vs {{ foo_bar_2_output }}`.

The codebase already treats this chaining as a bug: `_rewrite_error_code_mapping_refs_atomic` applies all substitutions in one sentinel-guarded pass precisely so `foo-bar -> foo_bar` and `foo_bar -> foo_bar_2` can't combine, and there's a regression test for it. But that guard only covers `error_code_mapping` — blocks, parameters, and `workflow_system_prompt` still get the per-entry loop. This runs on `POST /workflows` and `POST /workflows/{id}` whenever a `yaml_definition` is sent.

PR incoming that extends the existing atomic approach to the other three rewrite targets.

Separate but related: `replace_jinja_reference` only matches a name directly after `{{`, so references inside `{% if %}` / `{% for %}` or mid-expression (like the documented while_loop pattern `{{ current_index < max_attempts }}`) are never rewritten at all when their identifier is renamed. That's a matching-coverage question rather than a chaining one, so I've left it out of the PR — happy to file it separately.

MatthewSynthia:
Fixes #7554

## Description

`sanitize_workflow_yaml_with_references` rewrites Jinja references once per renamed identifier, sequentially. When collision-driven renames overlap — `user email` -> `user_email` while the pre-existing `user_email` gets suffixed to `user_email_2` — the later substitution re-hits the earlier one's output, so a just-repaired reference chains on to the wrong parameter: `Send to {{ user email }} and CC {{ user_email }}` comes out as `Send to {{ user_email_2 }} and CC {{ user_email_2 }}`. Same thing for block labels via the `{label}_output` rewrites, which can collapse two different blocks' output references onto one block. Nothing errors at import or run time; the workflow just runs with the wrong values.

The codebase already solves this exact problem for `error_code_mapping`: `_rewrite_error_code_mapping_refs_atomic` applies all substitutions in a single sentinel-guarded pass so `foo-bar -> foo_bar` and `foo_bar -> foo_bar_2` can't combine, with a regression test pinning it. This PR extends that approach to the other three rewrite targets:

- the sentinel logic moves into a shared `_build_atomic_jinja_rewriter`; the error_code_mapping helper now uses it unchanged
- steps 3/4 build one ordered substitution list (`{label}_output` before the bare label, mirroring the error_code_mapping ordering) and rewrite blocks, parameters, and `workflow_system_prompt` in one atomic pass via `_rewrite_jinja_refs_atomic`
- the per-entry `_replace_references_in_value` had no remaining callers and is removed

The direct-string pass (`source_parameter_key` etc.) is intentionally left as-is — its chaining problem is scoped to #7278.

## How Has This Been Tested

- New `tests/unit/test_workflow_sanitizer_jinja_rewrites.py` covers param-key chaining in block fields, label `_output` chaining, system-prompt/param-default chaining, and pins non-colliding renames + lookalike identifiers as untouched. The three chaining tests fail on `main` and pass with the fix; the pin test passes on both.
- Drove the sanitizer through a 9-case before/after panel (5 broken collision cases, 4 unchanged-behavior pins): main resolves 4/9, the fix 9/9, and the pins pass on both revisions — the no-behavior-change evidence for imports without collisions.
- Full suite green: `uv run pytest` (16,649 passed, 84 skipped, 1 xfailed); pre-commit ruff / ruff-format / isort / mypy / pyupgrade / autoflake all pass.

---
*Disclosure: prepared with AI assistance (Claude); I reviewed the diff, the root-cause trace against the existing error_code_mapping guard, and the local test runs before opening.*

### DIFF
--- a/skyvern/schemas/workflows.py
+++ b/skyvern/schemas/workflows.py
@@ -2,7 +2,7 @@ import abc
 import functools
 from dataclasses import dataclass, field
 from enum import StrEnum
-from typing import Annotated, Any, Literal
+from typing import Annotated, Any, Callable, Literal
 
 import structlog
 from pydantic import BaseModel, Field, field_serializer, field_validator, model_validator
@@ -74,26 +74,12 @@ def _get_text_prompt_model_name_by_llm_key() -> dict[str, str]:
     return reverse_mapping
 
 
-def _replace_references_in_value(value: Any, old_key: str, new_key: str) -> Any:
-    """Recursively replaces Jinja references in a value (string, dict, or list)."""
-    if isinstance(value, str):
-        return replace_jinja_reference(value, old_key, new_key)
-    elif isinstance(value, dict):
-        return {k: _replace_references_in_value(v, old_key, new_key) for k, v in value.items()}
-    elif isinstance(value, list):
-        return [_replace_references_in_value(item, old_key, new_key) for item in value]
-    return value
-
-
-def _rewrite_error_code_mapping_refs_atomic(mapping: Any, substitutions: list[tuple[str, str]]) -> Any:
-    """Atomically rewrites Jinja references (keys and values) in an error_code_mapping dict.
+def _build_atomic_jinja_rewriter(substitutions: list[tuple[str, str]]) -> Callable[[str], str]:
+    """Builds a rewriter that applies all Jinja-reference substitutions to a string in one pass.
 
     Uses unique sentinels so chained rewrites don't occur when one substitution's new
     identifier matches another's old identifier (e.g. foo-bar -> foo_bar AND foo_bar -> foo_bar_2).
     """
-    if not isinstance(mapping, dict) or not substitutions:
-        return mapping
-
     sentinels = [(old, new, f"\x00SKY_SANITIZE_{i}\x00") for i, (old, new) in enumerate(substitutions)]
 
     def _apply(text: str) -> str:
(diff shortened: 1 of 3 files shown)


LawyZheng (changes requested):
The atomic Jinja rewrite itself looks sound, but this PR is not safe to merge standalone on head `0e2ec8ae6e215c8a1ee51acc0bc7ef755ddcbc61`.

Using the collision from the issue, the Jinja text is corrected to `{{ user_email }}` / `{{ user_email_2 }}`, but the final parameter keys are still `['user_email_2', 'user_email_2']`, and a `source_parameter_key` for the first parameter is rewritten to `user_email_2`. That leaves `{{ user_email }}` with no backing parameter, so the end-to-end workflow remains invalid even though the new Jinja assertions pass.

The adjacent direct-string fix is in #7278, but #7555 does not currently contain it, and the two heads have a content conflict in `skyvern/schemas/workflows.py`. Please land/restack onto #7278 (or include the equivalent direct-reference fix), resolve the conflict, and add regression assertions for the final parameter keys and `source_parameter_key`, not only the rendered Jinja strings.

Focused sanitizer tests pass (`77 passed`) and Ruff is clean; the missing end-to-end assertions are the blocker.

MatthewSynthia:
Restacked onto #7278 — you're right that these are two halves of the same bug. Its direct-string pass is what stops the sanitized `key` field from getting rewritten into a duplicate, and the atomic Jinja rewrite here sits on top of that.

On your repro the keys now come out `['user_email', 'user_email_2']`, `source_parameter_key` follows its target to `user_email_2`, and the Jinja lines up too — no dangling `{{ user_email }}`. I added a test that pins all three in one collision (final keys, `source_parameter_key`, and the rendered Jinja) so they can't drift apart again.

It'll show as behind until #7278 lands; I'll rebase onto main as soon as it does.

MatthewSynthia:
@LawyZheng ready for another look when you have time. I restacked onto #7278 like you suggested, so this now sits on top of its direct-string pass and just handles the chained-rewrite case.

PR #7560, opened 2026-07-19
[templating] Rewrite renamed Jinja identifiers inside {% %} statements and mid-expression

### LINKED ISSUE #7559: Renamed workflow identifiers are never rewritten inside {% %} statements or mid-expression Jinja references
### What happened

`replace_jinja_reference` in `skyvern/utils/templating.py` only matches an identifier immediately after `{{`:

```python
pattern = rf"\{{\{{(\s*){escaped_old_key}(?![a-zA-Z0-9_])"
```

So when the workflow import sanitizer (`sanitize_workflow_yaml_with_references`, runs on `POST /workflows` and `POST /workflows/{id}` with a `yaml_definition`) renames a parameter key or block label, references are only repaired when they sit at the head of a `{{ ... }}` expression. Two documented usage shapes are silently skipped:

* **Mid-expression references**: `{{ current_index < max attempts }}` — the exact shape of a while-loop `Condition`, which the docs describe as a Jinja template with the reserved `{{ current_index }}` counter (`docs/cloud/building-agents/configure-blocks.mdx`). `current_index` is at the head; the parameter being compared against never is.
* **Statement references**: anything inside `{% if ... %}` / `{% for ... %}` is never matched at all, including `{label}_output` guards like `{% if my-block_output.success %}`.

### Impact

After a rename (invalid identifier or a collision-driven `_2` suffix), the skipped references keep pointing at the old name. With default lax templating nothing errors at import or run time — conditions evaluate against undefined/wrong values, so loops mis-terminate and guards mis-fire. When the rename was collision-driven, the stale reference can silently bind to the *other* parameter that now owns the old name.

### Reproduce

```python
from skyvern.utils.templating import replace_jinja_reference

replace_jinja_reference("{{ current_index < max_attempts }}", "max_attempts", "max_attempts_2")
# returns unchanged; expected "{{ current_index < max_attempts_2 }}"

replace_jinja_reference("{% if my_flag %}go{% endif %}", "my_flag", "my_flag_2")
# returns unchanged; expected "{% if my_flag_2 %}go{% endif %}"
```

The same matcher backs every rewrite path in the sanitizer, including the sentinel-guarded `_rewrite_error_code_mapping_refs_atomic`, so all of them share the coverage gap.

### Notes

This is the matching-coverage defect noted at the end of Skyvern-AI/skyvern#7554 (scoped out of [#7555](<https://github.com/Skyvern-AI/skyvern/issues/7555>), which fixes the separate chaining problem). PR incoming that widens the matcher to full-token occurrences inside `{{ ... }}` and `{% ... %}` spans while leaving prose outside Jinja delimiters, attribute accesses, quoted string literals, and longer identifiers untouched.

dev-synthia:
Fixes #7559

## Description

`replace_jinja_reference` matched an identifier only immediately after `{{` (`\{\{(\s*)key(?!...)`), so every rewrite path in the workflow import sanitizer — blocks, parameters, `workflow_system_prompt`, and the sentinel-guarded `error_code_mapping` pass — silently skipped references that don't sit at the head of an expression. When a parameter key or block label is renamed at import (invalid identifier, or a collision-driven `_2` suffix):

- `{{ current_index < max_attempts }}` — the documented while-loop condition shape — keeps its stale right-hand side
- `{% if my-block_output.success %}` / `{% for x in items %}` statements are never touched at all

With default lax templating nothing errors; conditions evaluate against undefined or wrong values, and after a collision rename the stale reference can silently bind to the other parameter that now owns the old name.

The fix widens the matcher to full-token occurrences inside closed `{{ ... }}` and `{% ... %}` spans:

- token boundaries exclude attribute accesses (`foo.oldKey`), quoted string literals (`'oldKey'`), and longer identifiers (`oldKeyExtended`); a trailing `.`/`[` stays allowed so `{{oldKey.field}}` / `{{oldKey[0]}}` keep their access path with only the root renamed
- text outside Jinja delimiters is never rewritten (a new pin the old implementation happened to satisfy only for non-leading text)
- the legacy leading-position rewrite is kept as a fallback for unclosed `{{ oldKey` spans, so the new behavior is a strict superset of the old — all 8 pre-existing parametrized cases pass unchanged

No callers change: the sanitizer's per-target passes and `_rewrite_error_code_mapping_refs_atomic`'s sentinel substitution both go through this function and pick up the widened coverage. This is orthogonal to #7555 (chaining vs. matching coverage); they compose in either merge order.

## How Has This Been Tested

- 11 new parametrized cases on `replace_jinja_reference` (mid-expression comparison, repeated occurrences in one expression, `{% if %}`, `{% for %}`, whitespace-control `{%- -%}`, multiline expressions, plus unchanged-behavior pins for statement partial matches, attribute positions, string literals, prose outside delimiters, and unclosed-span leading rewrites) and 2 end-to-end tests through `sanitize_workflow_yaml_with_references` (while-loop `complete_criterion` with a renamed parameter; `{label}_output` guard inside `{% if %}`).
- The 7 fix-sensitive tests fail on `main` and pass with the fix; the 71 pre-existing tests in the file pass on both revisions.
- Full suite green: `uv run pytest` (16,705 passed, 84 skipped, 1 xfailed); pre-commit ruff / ruff-format / isort / mypy / pyupgrade / autoflake all pass.

---
*Disclosure: prepared with AI assistance (Claude); I reviewed the diff, the boundary-rule design, and the local test runs before opening.*

### DIFF
--- a/skyvern/utils/templating.py
+++ b/skyvern/utils/templating.py
@@ -8,8 +8,48 @@ class Constants:
     MissingVariablePattern = var_pattern = r"\{\{\s*([a-zA-Z_][a-zA-Z0-9_.\[\]'\"]*)\s*\}\}"
 
 
-# Closed Jinja expression/statement spans. DOTALL because expressions may span lines.
-_JINJA_SPAN_PATTERN = re.compile(r"\{\{.*?\}\}|\{%.*?%\}", re.DOTALL)
+# Characters that may precede a full-token occurrence of a key (identifier chars would
+# make it a longer identifier; a dot would make it an attribute access like foo.key).
+_TOKEN_BOUNDARY_BEFORE = set("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_.")
+# Characters that may not follow a full-token occurrence (identifier continuation).
+_TOKEN_BOUNDARY_AFTER = set("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_")
+
+
+def _rewrite_span_tokens(span: str, old_key: str, new_key: str) -> str:
+    """Rewrites full-token occurrences of old_key in one Jinja span, skipping string literals.
+
+    Single left-to-right pass. Quoted regions ('...' or "...", honoring backslash escapes)
+    are copied verbatim, so a key embedded anywhere inside a string literal is never touched.
+    """
+    parts: list[str] = []
+    i = 0
+    n = len(span)
+    key_len = len(old_key)
+    while i < n:
+        if span.startswith(old_key, i):
+            prev = span[i - 1] if i > 0 else ""
+            nxt = span[i + key_len] if i + key_len < n else ""
+            if (not prev or prev not in _TOKEN_BOUNDARY_BEFORE) and (not nxt or nxt not in _TOKEN_BOUNDARY_AFTER):
+                parts.append(new_key)
+                i += key_len
+                continue
+        char = span[i]
+        parts.append(char)
+        i += 1
+        if char in "'\"":
+            # Copy the whole string literal verbatim (backslash escapes included) so
+            # embedded occurrences of the key are never rewritten.
+            quote = char
+            while i < n:
+                literal_char = span[i]
+                parts.append(literal_char)
+                i += 1
+                if literal_char == "\\" and i < n:
+                    parts.append(span[i])
+                    i += 1
+                elif literal_char == quote:
+                    break
+    return "".join(parts)
 
 
 def replace_jinja_reference(text: str, old_key: str, new_key: str) -> str:
@@ -20,8 +60,12 @@ def replace_jinja_reference(text: str, old_key: str, new_key: str) -> str:
     {{ other < oldKey }}, {% if oldKey %}, {% for x in oldKey %}.
 
     Left untouched: occurrences outside Jinja delimiters, attribute accesses (foo.oldKey),
-    quoted string literals ('oldKey'), and longer identifiers that merely contain the key
-    (oldKeyExtended).
+    anything inside quoted string literals ('...oldKey...'), and longer identifiers that
+    merely contain the key (oldKeyExtended).
+
+    The scan is a single left-to-right pass: each span search resumes where the previous
+    span ended, so malformed input (e.g. thousands of unmatched braces) stays linear
+    instead of rescanning to the end from every opener.
 
     Args:
         text: The text to search in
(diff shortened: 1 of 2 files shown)


skyvern/utils/templating.py
```diff
@@ -21,13 +31,22 @@ def replace_jinja_reference(text: str, old_key: str, new_key: str) -> str:
...
-    # which prevents matching {{keyOther}} when searching for {{key}}
-    # Capture whitespace after {{ to preserve formatting (e.g., "{{ key }}" stays "{{ newKey }}")
     escaped_old_key = re.escape(old_key)
-    pattern = rf"\{{\{{(\s*){escaped_old_key}(?![a-zA-Z0-9_])"
-    return re.sub(pattern, rf"{{{{\1{new_key}", text)
+    # A full-token occurrence: not preceded by an identifier character, a dot (attribute
+    # access), or a quote (string literal), and not followed by an identifier character or
+    # a quote. A trailing dot/bracket stays allowed so {{oldKey.field}} and {{oldKey[0]}}
+    # keep their access path with only the root renamed.
+    token_pattern = re.compile(rf"(?<![a-zA-Z0-9_.'\"]){escaped_old_key}(?![a-zA-Z0-9_'\"])")
```
LawyZheng: **P2 — quoted string literals are still mutated.** The lookarounds only protect a key when it is immediately adjacent to a quote. A standalone token inside a longer literal is surrounded by spaces and is rewritten: `{{ notify('ping old_key now') }}` becomes `{{ notify('ping new_key now') }}`. The base implementation leaves this unchanged, so this is a new semantic regression and contradicts the stated `quoted string literals ... left untouched` contract. Please make the span rewrite quote-aware (or otherwise lexer-aware) and add negative tests for embedded tokens in both single- and double-quoted literals.

skyvern/utils/templating.py
```diff
@@ -8,10 +8,20 @@ class Constants:
     MissingVariablePattern = var_pattern = r"\{\{\s*([a-zA-Z_][a-zA-Z0-9_.\[\]'\"]*)\s*\}\}"
 
 
+# Closed Jinja expression/statement spans. DOTALL because expressions may span lines.
+_JINJA_SPAN_PATTERN = re.compile(r"\{\{.*?\}\}|\{%.*?%\}", re.DOTALL)
```
LawyZheng: **P2 — malformed delimiters cause quadratic scanning.** This lazy DOTALL span regex rescans to the end from every unmatched `{{`. On this exact head, inputs containing 1k/2k/4k/8k/16k unmatched braces (with `old_key` present) took approximately 0.002/0.008/0.030/0.121/0.476 seconds respectively—about 4× time for each 2× input. The helper is called once per rename substitution across imported workflow fields, so crafted YAML can multiply the cost. Please use a bounded/linear span scanner, or prove and enforce a sufficiently small hard input bound, and add a malformed-input performance regression test.

LawyZheng (changes requested):
The target behavior is valid and the focused tests pass, but this head has two blocking correctness/robustness issues in the new regex-based span rewrite. Please address the two inline findings, add the corresponding negative/performance regression tests, and rebase onto current `main` before re-review.

Related but non-blocking for this PR: sequential chaining in the other sanitizer passes is tracked separately by #7555.

dev-synthia:
Thanks for the review — both findings were real, and the measurements on the second one were appreciated.

- **Quoted literals**: the lookarounds were the wrong tool; they only protected tokens touching a quote. Replaced the span rewrite with a single-pass scanner that copies string literals verbatim (backslash escapes included), so embedded keys can't be rewritten anywhere in a literal. Added negative tests for single/double quotes in both expression and statement positions, plus escaped quotes.
- **Quadratic scanning**: replaced the lazy DOTALL regex with a left-to-right scan where each span search resumes where the previous span ended — malformed input is now linear. Unclosed spans keep the legacy leading-position rewrite (anchored pattern, no wildcards). Added a malformed-delimiter perf regression test (64k unmatched openers, generous 2s bound — the previous implementation takes ~30s there, the new one milliseconds).

Pushed as a separate commit on top for easier delta review, and rebased onto current main. Full suite green (16,453 passed), pre-commit clean.

dev-synthia:
Follow-up on the scanner commit: I noticed the unclosed-opener path curtailed coverage — after an unclosed `{{` or `{%`, the rest of the input only got the legacy leading-position rewrite, so a well-formed span later in the text (e.g. `{{ oops {% if old_key %}`) lost the mid-expression rewrite the lazy-regex version still had on that shape. That quietly broke the "strict superset of the old behavior" claim in the PR description.

Fixed in 896d199: an unclosed opener now gets its leading-position rewrite in place and the scan continues past it. To keep malformed input linear, opener positions are cached until the scan passes them, and a closer type that's absent from the rest of the text is never searched for again — so every `find()` covers a distinct stretch of input. Re-checked your adversarial shapes plus unclosed-openers-interleaved-with-closed-spans: all ~30ms at 64k openers.

Added parity tests for closed spans after an unclosed opener of either type (both fail on the previous commit). Full suite green (16,456 passed), pre-commit clean.

skyvern/utils/templating.py
```diff
@@ -8,10 +8,65 @@ class Constants:
...
 
 
+# Characters that may precede a full-token occurrence of a key (identifier chars would
+# make it a longer identifier; a dot would make it an attribute access like foo.key).
+_TOKEN_BOUNDARY_BEFORE = set("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_.")
+# Characters that may not follow a full-token occurrence (identifier continuation).
+_TOKEN_BOUNDARY_AFTER = set("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_")
+
+
+def _rewrite_span_tokens(span: str, old_key: str, new_key: str) -> str:
```
AronPerez, line 18: Interesting, is there a solution that already might exists within the Jinja dep to address this?

dev-synthia: Good question — I'd say that there are three things that rule out Jinja here: (1) the keys being renamed are usually invalid Jinja identifiers, so Jinja's lexer/AST can't see the old keys as single tokens; (2) Jinja has no source-to-source rewrite path — env.parse() doesn't round-trip back to template text; and (3) imported templates can be malformed, and env.lex/parse raise TemplateSyntaxError on unclosed delimiters while we want to degrade gracefully. The scanner here is a mini-lexer for the subset that matters: span boundaries and string literals.

PR #7580, opened 2026-07-20
feat(SKY-12490): clean up browser profile folders orphaned by hard-killed local MCP servers

marcmuon:


AronPerez:
Rebased onto current `main` and resolved the shared `skyvern-frontend/src/api/types.ts` conflict as a union, retaining both the eval timing/summary fields and mainline runtime-cost types. Current main already contains the SKY-12490 backend/tests and shared time-format changes, so the duplicate sync commits dropped during rebase.

The prior CI failure was the unrelated CDP discovery timeout; current main removed that proxy suite in #7583. During the requested SKY-12490 validation, I found and fixed a separate profile-cleanup race: after forced termination, the deletion helper could return before the child was reaped when the work deadline had expired. The fix grants the killed child the existing bounded stop reserve, with deterministic and real-process regression coverage.

Validation:
- `uv run pytest tests/unit/test_local_browser_profile.py tests/unit/test_run_commands_cleanup.py tests/unit/test_quickstart_command.py tests/unit/test_sdk_run_metadata.py -q` — 142 passed
- `npx vitest run src/util/timeFormat.test.ts` — 19 passed
- `uv run alembic heads` — one head (`e4db575f75ee`)
- `uv run pre-commit run --all-files` — passed
- `skyvern/forge/sdk/api/files.py` remains unchanged with the realpath + startswith sanitizer

Pushed `0498d8c15`.

PR #7582, opened 2026-07-20
DB-back the Odysseys runtime/tokens/cost panel (drop the hardcoded dataset)

trevor-cheung:


trevor-cheung:

Source PR: https://github.com/Skyvern-AI/skyvern-cloud/pull/13668

AronPerez:
Closing as redundant with evidence (same precedent as #6394).

Root cause of the red required check: the sync generator added migration a3a29ad2ab68 for eval_runtime_cost_results, but the OSS tree contains no corresponding SQLAlchemy model, route, or non-migration reference. Alembic therefore upgraded the table and then correctly proposed removing the table plus both indexes. This was a cloud-only migration carryover, so I removed it instead of inventing an unused OSS model.

Evidence after merging origin/main (never rebased):
- Pushed head: 1171d7b0adc0d83afcc2fc41589cf9da5f32a8cd
- The 11 original non-migration PR files are 11/11 blob-identical to origin/main, proving all synced CLI/library/frontend/test symbols already landed there.
- git diff --exit-code origin/main HEAD is empty, and origin/main is an ancestor of HEAD.
- uv run --extra server --group dev alembic heads: exactly one head, e4db575f75ee.
- Disposable PostgreSQL uv run ./run_alembic_check.sh: full upgrade passed; No new upgrade operations detected.
- npx vitest run src/util/timeFormat.test.ts: 1 file passed, 19/19 tests.
- skyvern/forge/sdk/api/files.py is byte-identical to main and retains the realpath + startswith(allowed_dir + os.sep) containment sanitizer.

There is no unique OSS change left to merge, so closing rather than landing an empty PR.

PR #7689, opened 2026-07-23
Keep real passwords out of generated login code, and stop endless safety-check retries (SKY-12866)

andrewneilson:

PR #7688, opened 2026-07-23
SKY-12733 Stop correct “no” answers from failing completed workflows

andrewneilson:

PR #7690, opened 2026-07-23
fix(SKY-12641): give the LLM clear no-file-download feedback

LawyZheng:


AronPerez:
@LawyZheng @andrewneilson Holding: CodeQL flags the same **alert #332** (`py/weak-sensitive-data-hashing`, high, `code_block_synthesis.py:784`) that hit #7688 and #7689. Pattern note: this is the **third consecutive mirror snapshot** carrying the pre-fix fingerprint code — andrewneilson's fix (`cc098da5c` on #7688, "Restore: keyed row-evidence fingerprint") appears to exist only on the #7688 branch, not in the upstream private repo, so every new sync cut will re-introduce this alert until the fix lands upstream. Suggested durable fix: sync cc098da5c's 2-file change (code_block_synthesis.py + test_copilot_code_block_synthesis.py) into upstream. Until then each mirror PR needs the same on-branch restore. CodeQL is required, so auto-merge stays blocked here meanwhile. — sync-PR manager

PR #7697, opened 2026-07-23
MCP-created code blocks default to the new code block editor experience

AronPerez:

PR #7698, opened 2026-07-23
Deduplicate the MCP block walk and store empty code block goals as NULL

AronPerez:


AronPerez:
Pushed take-main fingerprint restore (11th consecutive cut with the pre-fix sha256 `code_block_synthesis.py` — reverts main's keyed-scrypt fix, CodeQL alert #332; same as #7689/#7691-#7697). All other delta preserved. Durable fix: sync cc098da5c into the upstream private repo. — sync-PR manager

PR #7699, opened 2026-07-23
SKY-12916 Let Copilot safely continue through login and verification pages

andrewneilson:

PR #7700, opened 2026-07-23
Log the copilot terminal render decision per turn (SKY-12853)

pedrohsdb:


AronPerez:
Pushed take-main fingerprint restore (13th consecutive cut with the pre-fix sha256 `code_block_synthesis.py` — reverts main's keyed-scrypt fix, CodeQL alert #332; same as #7689/#7691-#7699). All other delta preserved. Durable fix: sync cc098da5c into the upstream private repo. — sync-PR manager

PR #7703, opened 2026-07-24
Save the workflow instead of giving up when Copilot already found the value (SKY-12907)

andrewneilson:


AronPerez:
Pushed take-main fingerprint restore (15th consecutive cut with the pre-fix sha256 `code_block_synthesis.py` — reverts main's keyed-scrypt fix, CodeQL alert #332; same as #7689/#7691-#7701). All other delta preserved. Durable fix: sync cc098da5c into the upstream private repo. — sync-PR manager

PR #7713, opened 2026-07-24
SKY-12933 Copilot: build the login step from what the scout demonstrated

andrewneilson:


AronPerez:
Pushed `bb9c38d40`: re-applied the keyed-fingerprint fix (cc098da5c) via **surgical 3-way patch** — this cut's `code_block_synthesis.py` carries new upstream content (function shifted to L870), so the usual whole-file take-main restore would have wiped that genuine delta. The 3-way apply landed cleanly on both files; scrypt verified present; all new upstream content preserved. CodeQL alert #332 should clear. Durable fix remains syncing cc098da5c upstream. — sync-PR manager

PR #7781, opened 2026-07-27
Stop re-injecting domUtils.js before retrying the injection itself (SKY-13012)

AronPerez:


AronPerez:
Closing as empty-redundant (precedent #6394). Verified against the full redundancy gate, not net-diff alone:

- **Net diff is empty**: `git diff origin/main...HEAD` = 0 files; GitHub's own `files` API also reports 0 files changed. The branch tree is identical to its merge-base tree, so this PR introduces nothing.
- **Not a false-empty**: this PR is BEHIND/MERGEABLE, not CONFLICTING, so no `-X theirs` conflict resolution is involved — the empty result is unambiguous.
- **Substance is on main**: the SKY-13012 work (stop re-injecting domUtils.js before retrying the injection itself) already landed via a cumulative sibling. `skyvern/webeye/utils/page.py:729` carries the fix and its explanatory comment — "The bootstrap call already IS the domUtils.js injection, so the retry below…" — alongside the context-loss re-injection path at :721. Nothing is lost by closing.

Reopen if the mirror intended additional content here that never got applied.

PR #7905, opened 2026-07-31
SKY-13244 Select a connected email account as the 2FA identifier (re-land)

suchintan:


suchintan:

Source PR: https://github.com/Skyvern-AI/skyvern-cloud/pull/14233

PR #7912, opened 2026-07-31
fix: hold a reference to the store_artifact task

noron12234:
## What

`update_script_files` overwrites a script file in storage by spawning the write and throwing the task away:

```python
artifact = await app.DATABASE.artifacts.get_artifact_by_id(script_file.artifact_id, organization_id)
if artifact:
    # override the actual file in the storage
    asyncio.create_task(app.STORAGE.store_artifact(artifact, content_bytes))
```

[The event loop only keeps a weak reference to a running task](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task):

> Important: Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn't referenced elsewhere may get garbage collected at any time, even before it's done.

So the write can be collected before it lands. The function returns successfully either way, so the caller believes the file was updated — the next read just serves the old content. No exception, no log line.

Note this is the update path specifically: the `else` branch below it *awaits* `create_script_file_artifact`, so a newly created file is written synchronously while an overwrite is not.

## Why this reads as an oversight

The repo already does this correctly in three other places:

| file | line |
|---|---|
| `webeye/default_persistent_sessions_manager.py` | 193 — `_background_tasks: set[asyncio.Task[None]]`, used at 389-390 |
| `webeye/real_browser_state.py` | 99 — `_detached_teardown_tasks: set[asyncio.Task[None]]` |
| `webeye/browser_factory.py` | 193 — `path_task.add_done_callback(...)` |

## The change

Same pattern, module-level here since `update_script_files` is a module function rather than a method:

```python
_STORE_ARTIFACT_TASKS: set[asyncio.Task] = set()
...
_store_task = asyncio.create_task(app.STORAGE.store_artifact(artifact, content_bytes))
_STORE_ARTIFACT_TASKS.add(_store_task)
_store_task.add_done_callback(_STORE_ARTIFACT_TASKS.discard)
```

`discard` rather than `remove` so a double callback cannot raise. The set stays bounded by the number of in-flight writes.

I kept the write fire-and-forget rather than awaiting it — turning it into an await would change the latency of `update_script_files`, and that is a behaviour decision rather than a bug fix. Happy to do that instead if you'd prefer the write to be synchronous like the create path.

## Verification

```console
$ ruff check skyvern/services/script_service.py       # ruff 0.14.1, pinned in .pre-commit-config.yaml
All checks passed!
$ ruff format --check skyvern/services/script_service.py
1 file already formatted
$ python -m py_compile skyvern/services/script_service.py
```

No test: the failure is a garbage-collection race, so a test would have to force a GC at a chosen moment and assert a task did *not* vanish — flaky by construction. Happy to add one if you have a shape in mind.

Found with an AST scan for `create_task` / `ensure_future` results discarded as bare expression statements. There are ten other hits in the repo (streaming channels, workflow blocks, agent speculation); those spawn UI/streaming work rather than storage writes, so I left them out to keep this diff to one concern. Happy to follow up if you want them covered.


noron12234:
Closing this. I am cleaning up the PRs I have open and keeping only the ones a maintainer has already picked up; nobody has needed to look at this one, and leaving it open just adds to your queue.

The branch stays on my fork, so nothing is lost if the fix turns out to be worth having. Sorry for the noise.

PR #7956, opened 2026-08-01
feat: tolerate missing local LLM setup with diagnostics and setup banner

suchintan:


AronPerez:
#6394 redundancy proof: merged current origin/main locally with git merge origin/main, never rebase. After resolving Settings.test.tsx and handler.py to current main behavior, git diff --quiet origin/main -- exited 0, so the final merge tree is byte-for-byte identical to main. Every feature file and added public identifier is already on main: LLMSetupBanner, useLLMDiagnostics, frontend and backend LLMDiagnosticsResponse, evaluate_llm_status, llm_status, the custom-llms settings deep-link, internal_llms.py, and test_internal_llms.py. The containing sync reached main through #7958 at commit 3e4838371d22fedb70d71a8cae9cb723b3479ac0. The stale-head CI failure was the missing workflows.mask_secrets migration; current main includes it, and uv run alembic heads reports the single head e1ffa32660cc. Focused verification passed: Settings Vitest 1 of 1 and LLM diagnostics pytest 6 of 6. The live PR reports 70 cumulative diff files while the resolved final tree reports 0, so this PR is redundant.

PR #7769, opened 2026-07-27
[Database] Fix: use portable distinct step count

Heman10x-NGU:
## Description

Replace the row-value `COUNT(DISTINCT ...)` task-step query with a portable distinct-pairs subquery followed by `COUNT(*)`. This preserves the existing task-ID and organization filters and lets SQLite enforce the max-step count.

## How Has This Been Tested?

`uv run --extra server --group dev pytest tests/unit/forge/sdk/db/test_tasks_repository.py -q`

The new SQLite-backed regression test covers repeated retries for one step order and confirms that the unique order count is returned successfully.

`ruff check` and `ruff format --check` passed for the changed files.

`pre-commit run --all-files` did not complete because the unrelated frontend hook could not find `lint-staged`; no browser, Docker, or E2E checks were run.

## Artifacts (if appropriate):

None.

PR #7963, opened 2026-08-02
fix: make unique step counting SQLite-compatible

### LINKED ISSUE #7763: SQLite (the default local DB) cannot run tasks: `row value misused` from count(distinct(tuple_(...)))
## Summary

Every task run fails on the **SQLite** backend — which is the out-of-the-box default — with:

```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) row value misused
[SQL: SELECT count(DISTINCT (steps.task_id, steps."order")) AS count_1
```

The task is marked `failed` with `failure_reason: "Database error when running task 2.0"`, which hides the real cause.

## Cause

`skyvern/forge/sdk/db/repositories/tasks.py:304`

```python
select(func.count(distinct(tuple_(StepModel.task_id, StepModel.order))))
```

with the comment above it at line 299:

```
Basically translate this sql query into a SQLAlchemy query: select count(distinct(s.task_id, s.order)) from steps s
```

`tuple_()` emits a SQL **row value**. Postgres accepts a row value inside `COUNT(DISTINCT ...)`; **SQLite does not**, and raises `row value misused`. So this query is Postgres-only, while `_default_database_string()` in `skyvern/config.py` points at SQLite:

```python
db_path = Path.home() / ".skyvern" / "data.db"
return f"sqlite+aiosqlite:///{db_path}"
```

whose docstring says it exists "so that `skyvern run server` works out of the box without Docker or Postgres."

## Reproduce

```bash
uv venv --python 3.13 .venv
uv pip install "skyvern[server,local]==1.0.47" --python .venv/bin/python
# .env with any working LLM provider; leave DATABASE_STRING unset (SQLite default)
./.venv/bin/skyvern run server
```

Then submit any task, e.g.:

```bash
curl -X POST http://127.0.0.1:8000/v1/run/tasks \
  -H "Content-Type: application/json" -H "x-api-key: $KEY" \
  -d '{"prompt":"Report the main heading text on this page.","url":"https://example.com","engine":"skyvern-2.0"}'
```

Result: `status: failed`, and the server log carries the traceback above. The server itself starts fine and bootstraps 54 tables — this is purely the query.

## Suggested fix

Count distinct pairs in a dialect-portable way, e.g. a subquery:

```python
subq = select(StepModel.task_id, StepModel.order).where(...).distinct().subquery()
select(func.count()).select_from(subq)
```

which works on both backends and avoids the row value entirely.

## Environment

- skyvern 1.0.47 (released wheel from PyPI)
- macOS arm64, CPython 3.13.14, `uv`
- SQLite default (`DATABASE_STRING` unset), 54 tables created successfully

## Related

Two other things found while getting to this, mentioned only for context:

- On Postgres, `skyvern run server` starts and migrates but seeds **no** organization or auth token (`organizations = 0`, `organization_auth_tokens = 0`), so the API returns 403 and is unusable unless the interactive `skyvern quickstart` is used — and that dies non-interactively with `EOF when reading a line`.
- The released wheel ships the `alembic` library but not the migration scripts, so a Postgres user has to run migrations from a source checkout.

Happy to open those separately if useful.

sardortaylakov:
## Description

Fixes #7763 by replacing the PostgreSQL-specific `COUNT(DISTINCT (task_id, order))` expression with a portable `COUNT(*)` over a `SELECT DISTINCT` subquery.

The two commits preserve the red/green progression:

1. Add a SQLite regression test covering duplicate retries, multiple task IDs, task filtering, and organization scoping.
2. Update the query so the same test passes on SQLite while preserving the existing count semantics.

Fixes #7763

## How Has This Been Tested?

```bash
uv run pytest tests/unit/forge/sdk/db/test_tasks_repository.py -q
uv run ruff check skyvern/forge/sdk/db/repositories/tasks.py tests/unit/forge/sdk/db/test_tasks_repository.py
uv run ruff format --check skyvern/forge/sdk/db/repositories/tasks.py tests/unit/forge/sdk/db/test_tasks_repository.py
```

Before the fix, the regression test failed with:

```text
sqlite3.OperationalError: row value misused
1 failed
```

After the fix:

```text
1 passed
All checks passed!
2 files already formatted
```

[Failing CI run on the regression-test commit](https://github.com/Skyvern-AI/skyvern/actions/runs/30729222948/job/91446371835)

### DIFF
--- a/skyvern/forge/sdk/db/repositories/tasks.py
+++ b/skyvern/forge/sdk/db/repositories/tasks.py
@@ -6,7 +6,7 @@ from decimal import Decimal
 from typing import Any, Sequence
 
 import structlog
-from sqlalchemy import and_, delete, distinct, func, select, tuple_, update
+from sqlalchemy import and_, delete, func, select, update
 
 from skyvern.forge.sdk.db._error_handling import db_operation
 from skyvern.forge.sdk.db.base_alchemy_db import read_retry
@@ -300,18 +300,17 @@ class TasksRepository(BaseRepository):
         task_ids: list[str],
         organization_id: str,
     ) -> int:
-        """
-        Get the total count of unique (step.task_id, step.order) pairs of StepModel for the given task ids
-        Basically translate this sql query into a SQLAlchemy query: select count(distinct(s.task_id, s.order)) from steps s
-        where s.task_id in task_ids
-        """
+        """Count unique ``(task_id, order)`` pairs for the given tasks."""
         async with self.Session() as session:
-            query = (
-                select(func.count(distinct(tuple_(StepModel.task_id, StepModel.order))))
+            distinct_steps = (
+                select(StepModel.task_id, StepModel.order)
                 .where(StepModel.task_id.in_(task_ids))
                 .where(StepModel.organization_id == organization_id)
+                .distinct()
+                .subquery()
             )
-            return (await session.execute(query)).scalar()
+            query = select(func.count()).select_from(distinct_steps)
+            return (await session.execute(query)).scalar_one()
 
     @db_operation("get_task_step_models")
     async def get_task_step_models(self, task_id: str, organization_id: str | None = None) -> Sequence[StepModel]:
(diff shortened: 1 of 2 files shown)


AronPerez:
Closing this in favor of https://github.com/Skyvern-AI/skyvern/pull/7769

PR #7973, opened 2026-08-02
fix: don't block the event loop while waiting for the browser to start

noron12234:
## What

`_create_cdp_connection_browser` is `async def`, but waits for the spawned browser with a synchronous sleep:

```python
browser_process = subprocess.Popen([...], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
# Add small delay to allow browser to start
time.sleep(1)
if browser_process.poll() is not None:
    raise Exception(f"Failed to open browser. browser_path: {browser_path}")
```

A blocking sleep inside a coroutine stalls the **entire event loop**, not just this call. For that second nothing else on the loop runs — every other task in the process is frozen, however unrelated.

`await asyncio.sleep(1)` waits the same wall-clock second while letting the loop keep working. `asyncio` is already imported in this file (line 3).

## Change

```diff
  # Add small delay to allow browser to start
- time.sleep(1)
+ await asyncio.sleep(1)
```

This was the only `time` usage in the file, so its import goes with it — that is the second line of the diff and the reason it is two lines rather than one.

No behaviour change beyond the loop staying responsive: the same delay, the same `poll()` check after it.

## Verification

```console
$ ruff check skyvern/webeye/browser_factory.py        # ruff 0.14.1, pinned in .pre-commit-config.yaml
All checks passed!
$ ruff format --check skyvern/webeye/browser_factory.py
1 file already formatted
$ python -m py_compile skyvern/webeye/browser_factory.py
```

Found with an AST scan for blocking calls (`time.sleep`, `requests.*`, `subprocess.run`, `urlopen`) inside `async def` bodies, excluding anything passed to `run_in_executor` / `to_thread` and anything in a nested sync `def`. This was the only hit in the repo.


AronPerez (approved):

noron12234:
Rebased onto current `main` — the PR was `BEHIND`, now `MERGEABLE` and approved. Ready to merge whenever you are.

PR #8163, opened 2026-08-08
feat: add error_code_mapping manifest + ErrorCode primitive to CodeBlock (SKY-13668)

AronPerez:


AronPerez:
Closing as redundant: all content is already present on main. Verified net-diff-empty (git diff --quiet origin/main..HEAD, exit 0) and symbols-on-main (337 identifier tokens across the three-dot additions, 0 absent from origin/main). No Alembic payload; files.py sanitizer verified intact (realpath + allowed_dir+os.sep startswith at 393-399, local containment at 567-572).

Issue #103, opened 2024-03-18
Refactor the setup


LawyZheng:
If I update the LLM provider now, my whole env will be re-setup again, including poetry install, database install, playwright install...

As the PR https://github.com/Skyvern-AI/skyvern/pull/102 said, I simply split the functions apart. Sometimes I just want to setup some parts of the project, such as `env`, or `database`(actually my db is setup in the remote server, but I can't just execute `alembic upgrade head` and `create secrets.toml` alone).

For the long term, the setup still should be refactored by Python scripts or CLI tools to deal with more complicated commands like `docker build` for image building, `clean` for local cache cleaning, `pytest` for testing, `uninstall` for project uninstalling...

ykeremy:
>For the long term, the setup still should be refactored by Python scripts or CLI tools to deal with more complicated commands like docker build for image building, clean for local cache cleaning, pytest for testing, uninstall for project uninstalling...

❤️❤️❤️ agreed

PR #8217, opened 2026-08-14
Add the selector engines Skyvern's producers emit to skycdp (SKY-14071)

wintonzheng:
Sync PR: 15140
Sync SHA: 636b919c9c9b3dbb53b0e9fda6d22401a3f508cd

### DIFF
--- a/skyvern/core/script_generations/generate_script.py
+++ b/skyvern/core/script_generations/generate_script.py
@@ -338,6 +338,7 @@ ACTION_MAP = {
     "upload_file": "upload_file",
     "select_option": "select_option",
     "goto": "goto",
+    "goto_url": "goto",
     "scroll": "scroll",
     "keypress": "keypress",
     "type": "type",
@@ -406,7 +407,7 @@ def _build_semantic_selector(act: dict[str, Any]) -> str | None:
     return None
 
 
-ACTIONS_OPT_OUT_INTENTION_FOR_PROMPT = ["extract"]
+ACTIONS_OPT_OUT_INTENTION_FOR_PROMPT = ["extract", "goto", "magic_link"]
 
 INDENT = " " * 4
 DOUBLE_INDENT = " " * 8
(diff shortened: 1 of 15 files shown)


AronPerez:
**Sync-queue triage: held on a real (but almost certainly benign) CodeQL alert — needs an author/security call.**

**State:** `MERGEABLE` / `BLOCKED`, head `3655300a`, merge-base == current `main` (`05f36a42`) — not behind, no `update-branch` needed.

**Branch-protection required checks are both green:** `Run tests and pre-commit hooks` ✅, `Frontend Lint and Build` ✅. Also green: `Analyze (actions/javascript-typescript/python)`, `pip Package Smoke Tests (3.11/3.13)`.

**What is actually blocking:** `auto-merge-sync.yml` ignores only `auto-merge` and `Mintlify Deployment`, and accepts only `success|skipped|neutral`. The `CodeQL` results check is `failure`, so the auto-merge job exits 1. This is *not* a branch-protection block — it is the auto-merge waiter.

**The alert** — code-scanning `#341`, `py/incomplete-url-substring-sanitization`, high, PR-introduced (baseline max high on `main` is `#103`):

`tests/unit_tests/webeye/test_navigation_retry.py:566`
```python
assert any("portal.example.com" in repr(entry) for entry in logs)
```

**Assessment:** this reads as a false positive. The line is a test assertion verifying `redact_url_secrets` keeps the host visible for diagnosability; it makes no sanitization decision and guards no control path. The rule fires on the syntactic shape `"<host>" in <str>`, which is only dangerous in product code. Precedent: `main` already carries the identical rule as an open, undismissed high alert (`#89`, `tests/unit_tests/test_url_validators_extended.py:63`, since 2025-12-14) — pre-existing alerts don't block, newly-introduced ones do.

**Not doing, deliberately:** no rerun (an AdvSec check is not a flake — it completed in 3s with a real verdict), no `--admin` merge past a security finding, and no self-dismissal of the alert.

**Options, author's call:**
1. Dismiss `#341` as *used in tests* — fastest, and consistent with how `#89` sits on `main`.
2. Reshape the assertion so the rule can't fire (e.g. compare a parsed `urlparse(...).netloc` instead of a literal substring). Note this is a mirror PR, so an OSS-only edit will drift from upstream and be re-mirrored — the durable version of this fix belongs upstream.
3. Add a CodeQL config excluding `tests/**` for this rule (no `.github/codeql*` config exists today) — repo-wide policy change.

Holding here until one of those lands. cc @wintonzheng

PR #8287, opened 2026-08-24
security: upgrade react-router-dom to 7.18.2

wintonzheng:
## Description

Upgrades `react-router-dom` from 6.30.4 to 7.18.2, closing three open Dependabot alerts on this repo (#902, #903, #904 - open redirect via backslash in `<Link>`/`useNavigate`, CVE-2026-53666 / 53668 / 53669). The v7 line is the only place these are fixed. Going to 7.18.2 rather than 7.18.0/7.18.1 also clears GHSA-qwww-vcr4-c8h2 (a CSRF bypass in the unstable RSC paths, which this app does not use).

Lockfile + manifest only - no source changes are needed, because the one v7-sensitive spot in this tree was already fixed upstream in skyvern-cloud and synced here: `RunRouter`'s legacy `blocks -> overview` redirect under the `runs/:runId/*` splat route uses an absolute path, which resolves correctly under both v6 relative-splat semantics and v7's.

For reviewers, the v7 defaults that change behavior and why they are safe here:
- **Relative splat paths** now resolve against the full matched URL. The splat routes in `src/router.tsx` are `runs/:runId/*` and `workflows/*`; the only relative redirect inside one is the RunRouter case above, already absolute.
- **Navigations run in `React.startTransition`.** No component in `src/` resets router state urgently alongside a navigation (the one case that did was in the cloud overlay, not synced).
- `json`, `defer`, `useFetcher`, `<Form>`, `fallbackElement` and `unstable_*` are unused, so the remaining v7 flags are no-ops here.

## How Has This Been Tested?

`npm install` resolves react-router and react-router-dom at 7.18.2. `tsc --noEmit` exits 0, `vitest run` passes 4305 tests across 445 files with 0 failures, and `vite build` produces a production bundle. The same upgrade was validated end to end on the cloud side (5264 tests, plus a dev-server boot confirming the splat redirect resolves to `/runs/<id>/overview`).

## How Has This Been Tested? (added: OSS-specific boot verification)

Answering "are we sure this isn't breaking?" directly - the earlier run proved the v7 behavior on the cloud entrypoint, so here is the same check on the OSS one (`src/main.tsx` + `src/router.tsx`, bundled by this repo's vite 6, not cloud's vite 8). Dev server booted on 7.18.2 and every route that depends on a changed v7 default was driven in a real browser:

| route | resolved to | what it proves |
|---|---|---|
| `/runs/wr_test/blocks` | `/runs/wr_test/overview` | the splat-relative redirect, the one spot v7 changed |
| `/runs/tsk_test` | `/runs/tsk_test/actions` | index redirect inside the `runs/:runId/*` splat |
| `/workflows/whatever` | `/agents/whatever/runs` | legacy `workflows/*` splat rewrite keeps its param |
| `/` | `/discover` | ordinary top-level redirect |

Zero router or React errors on any of the four (API calls 502 with no backend running, which is expected and unrelated). Plus `tsc --noEmit` clean, `vitest run` 4305 passed / 0 failed across 445 files, and `vite build`.

Worth noting how small the dependency change is: besides the two version lines, the lockfile only adds `cookie` and `set-cookie-parser` under react-router and drops `@remix-run/router` - v6's separate router core, which v7 absorbed into the `react-router` package. No transitive dependency is added to the app bundle.

OSS app booted on react-router 7.18.2 with the splat route resolving to /runs/wr_test/overview

Screenshot above - OSS dev boot on 7.18.2.


AronPerez (changes requested):
are we sure this isn't breaking?

wintonzheng:
@AronPerez fair question on the v6 → v7 jump — here's the compat audit rather than just "CI is green".

**Why the major bump is low-risk for this app specifically**

The app is already on the v6.4+ data router (`createBrowserRouter` + `RouterProvider` in `src/App.tsx` / `src/router.tsx`), which is the architecture v7 is built on. The v6 → v7 breaking changes fall into two buckets, and both check out:

1. **Removed APIs** — `json()` and `defer()` are gone in v7. Neither is imported from `react-router*` anywhere in the frontend.

2. **Former `future` flags becoming default** — these are the behavior changes, and almost all of them only affect loaders/actions/fetchers. This app uses none of that machinery:

   | API | Occurrences |
   |---|---|
   | `useLoaderData` | 0 |
   | `useActionData` | 0 |
   | `useFetcher` | 0 |
   | `useSubmit` | 0 |
   | route-level `loader:` | 0 |

   So `v7_fetcherPersist`, `v7_partialHydration`, and `v7_skipActionErrorRevalidation` are inert. `v7_normalizeFormMethod` is also inert: the ~428 `<Form` hits are react-hook-form / shadcn, not react-router — there are **0** imports of `Form` from `react-router-dom`.

**The one flag with a real surface: `v7_relativeSplatPath`**

Splat routes do exist — `runs/:runId/*` and `workflows/*` (plus `/sign-in/*`, `/sign-up/*` in the cloud router). Relative navigation inside them is narrow: `<Navigate to="overview">` and `<Navigate to="actions">` in `routes/runs/RunRouter.tsx`, both inside a descendant `<Routes>` tree, which re-bases on the parent splat match the same way in v6 and v7. `routes/runs/runViewTabsNavigation.test.tsx` covers exactly that tab navigation.

**Status:** `main` is merged in as of 82214693. Frontend Lint and Build, CodeQL, and both pip smoke suites are green; the test job is still running.

**Honest residual risk:** a green build proves compilation, not routing at runtime. The two things worth a manual click-through before merge are (a) run-detail tab switching under `runs/:runId/*` and (b) deep links into the workflow editor under `workflows/*` — those are the only places `v7_relativeSplatPath` could bite. Happy to run that smoke pass if you'd rather see it before approving.

AronPerez (changes requested):
hmm

AronPerez (approved):

PR #8294, opened 2026-08-24
Log page-analysis deadline exhaustion at warning consistently

AronPerez:
Sync PR: 15988
Sync SHA: 1426584fe096b16038b6b37d918ede86524416df

### DIFF
--- a/packages/skyvern-ui/pyproject.toml
+++ b/packages/skyvern-ui/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "skyvern-ui"
-version = "1.0.50"
+version = "1.0.51"
 description = "Prebuilt Skyvern UI assets for the Skyvern CLI"
 authors = [{ name = "Skyvern AI", email = "info@skyvern.com" }]
 requires-python = ">=3.11,<3.14"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "skyvern"
-version = "1.0.50"
+version = "1.0.51"
 description = ""
 authors = [{ name = "Skyvern AI", email = "info@skyvern.com" }]
 requires-python = ">=3.11,<3.15"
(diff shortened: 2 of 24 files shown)


AronPerez:
Sync-PR manager triage — **one real failure, needs the author. Not a flake, not infra.**

**`Run tests and pre-commit hooks` is red on this PR's own new test** (1 failed / 4324 passed):

```
FAIL src/routes/runs/runViewTabsNavigation.test.tsx
  > a relative target under the runs splat misroutes against the matched segment
AssertionError: expected '/runs/wr_1/output' to be '/runs/wr_1/overview/output'
  Expected: "/runs/wr_1/overview/output"
  Received: "/runs/wr_1/output"
  at runViewTabsNavigation.test.tsx:178:32
```

`runViewTabsNavigation.test.tsx` (+180) and `runViewTabBasePath.ts` (+21) are **new in this PR** — neither exists on `main` — so this is not a pre-existing or inherited failure. `runViewTabBasePath.test.ts` passes; only the navigation characterization test fails.

**Why I did not "fix" it:** the test name says the relative target *misroutes*, and it asserts the misrouted path `/runs/wr_1/overview/output`. React Router here produced `/runs/wr_1/output` — the path that looks *correct*. So the test encodes behavior that does not reproduce in open source. That reads as an OSS/cloud divergence in `SwitchBarNavigation` or the router version, not a typo. Editing the assertion would mask a real divergence, and changing the component would be a behavior change — either way the intent is yours, so I am routing rather than guessing.

**What I did do:** this PR arrived carrying the recurring three-file `aes.py` / `bootstrap.py` / `test_aes_fallback_decrypt.py` reversion (sha256 -> md5 normalization, plus deletion of `test_decrypts_ciphertext_created_with_legacy_primary_normalization`). I restored all three to `main`'s versions in `e1f4cb8f0` — 3 files, byte-identical to `main`, your 16-file payload untouched, 0 deletions. **No action needed from you on that part.**

Heads-up: that reversion has now arrived on 19 sync PRs in ~24h and is re-emitted by the mirror generator on essentially every one, so per-PR fixes do not stop it (#8288 reverted #8286's fix 90 seconds after it merged). The durable fix is to sync `main`'s versions of those three files into the private mirror source.

PR #8331, opened 2026-08-30
Task V3: hand the previous block's outcome and position to the next block (flagged, default off)

pedrohsdb:
Sync PR: 16359
Sync SHA: 7d4770f8ceded2647a2b347242238a9d25468d31


pedrohsdb:

Sync PR: 16359

PR #8399, opened 2026-09-02
Record why a browser session was closed so an aborted one is not counted as completed

AronPerez:
Sync PR: 16086
Sync SHA: e95c5976f2880db80cbf7f345c17e7169f9e4e09


AronPerez:

Sync PR: 16086

PR #8449, opened 2026-09-04
Grant the Estuary CDC role SELECT on three app tables missing from the warehouse mirror (SKY-15607)

marcmuon:
Sync PR: 16608
Sync SHA: ba227aab8f8bd9bc023fe3a5c7caf1e208a4b90b

### DIFF
--- a/alembic/versions/2026_09_04_0608-456c370b0f7d_grant_the_estuary_cdc_role_select_on_three_app_tables_missing_from_the_warehouse_mirror.py
+++ b/alembic/versions/2026_09_04_0608-456c370b0f7d_grant_the_estuary_cdc_role_select_on_three_app_tables_missing_from_the_warehouse_mirror.py
@@ -1,0 +1,64 @@
+"""grant the Estuary CDC role SELECT on three app tables missing from the warehouse mirror
+
+Revision ID: 456c370b0f7d
+Revises: ac35398557c1
+Create Date: 2026-09-04T06:08:50.733883+00:00
+
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "456c370b0f7d"
+down_revision: Union[str, None] = "ac35398557c1"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+_CDC_ROLE = "flow_capture"
+_PUBLICATION = "flow_publication"
+_MIRRORED_TABLES = (
+    "user_onboarding",
+    "recovery_guidance_experiment_assignments",
+    "workflow_tag_events",
+)
+
+
+def upgrade() -> None:
+    for table in _MIRRORED_TABLES:
+        op.execute(f"""
+        DO $$
+        BEGIN
+            IF NOT EXISTS (SELECT 1 FROM pg_roles WHERE rolname = '{_CDC_ROLE}') THEN
+                RAISE NOTICE '{_CDC_ROLE} role not found, skipping grant on {table} (expected on local/OSS)';
+                RETURN;
+            END IF;
+            IF to_regclass('public.{table}') IS NULL THEN
+                RAISE NOTICE 'public.{table} not found, skipping grant';
+                RETURN;
+            END IF;
+
+            GRANT SELECT ON public.{table} TO {_CDC_ROLE};
+
+            IF EXISTS (SELECT 1 FROM pg_publication WHERE pubname = '{_PUBLICATION}' AND NOT puballtables) THEN
+                IF NOT EXISTS (
+                    SELECT 1 FROM pg_publication_tables
+                    WHERE pubname = '{_PUBLICATION}' AND schemaname = 'public' AND tablename = '{table}'
+                ) THEN
+                    BEGIN
+                        ALTER PUBLICATION {_PUBLICATION} ADD TABLE public.{table};
+                    EXCEPTION WHEN insufficient_privilege THEN
+                        RAISE WARNING 'cannot alter {_PUBLICATION} as %; run manually as its owner: '
+                                      'ALTER PUBLICATION {_PUBLICATION} ADD TABLE public.{table};', current_user;
+                    END;
+                END IF;
+            END IF;
+        END $$;
+        """)
+
+
+def downgrade() -> None:
+    pass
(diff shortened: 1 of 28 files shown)


marcmuon:

Sync PR: 16608

AronPerez:
@marcmuon — flagging a **reliable** CI failure on this PR's head, with measurements, before it gets re-cut upstream.

## Summary

`tests/unit/test_agent_cleanup_download_suffix.py::test_cleanup_timeout_bounds_browser_download_handler_drain` fails **5/5** on head `2460d04e` in CI. This is not a flake.

| Sample | elapsed, vs the `< 0.2` bound |
|---|---|
| attempt 1 | 1.1056s |
| attempt 2 | 0.9986s |
| attempt 3 | 1.3598s |
| attempt 4 | 1.3547s |
| attempt 5 | 1.1174s |

All five are run `33911210507`, attempts 1–5, head unchanged throughout. Control: `origin/main` was **8/8 green** on the same suite over the same window (e.g. run `33890138386`, 25123 passed in 769.69s).

## What is established

- **The drain itself works.** The captured log shows `Timeout to save downloaded files` firing, with the remaining cleanup completing 0.5ms later. Only the wall-clock bound breaks.
- **It is not a constant.** `BROWSER_PAGE_CLOSE_TIMEOUT=5`, `SAVE_DOWNLOADED_FILES_TIMEOUT=180` (patched to `0.01` by the test). That 0.01s timer fired ~1s late, so the event loop was not servicing timers for ~1–1.4s.
- `asyncio.wait_for(..., timeout=0.5)` never raised, so `clean_up_task` completed — this is **loop blocking, not a hang**.
- Elapsed values vary (0.9986–1.3598s), so it is a **variable** stall rather than a fixed wait.
- The failing test *function* is byte-identical to `main`, and the `0.2` bound predates this PR (landed in #7651).
- Four symbols are new here and absent from `main`: `snapshot_recording_prefixes`, `_close_claimed_download_popups`, `_close_credited_download_popups`, `stream_artifact_prefix_from_path`.

## What is *not* established

The mechanism. It does not reproduce locally: 22/22 isolated runs, the full 37-test file 3×, and the full suite 2/2 all pass on this head. The dev box runs the suite in 330–370s against CI's 660–770s, so it likely never starves the loop the way the CI runner does. We did not isolate which change causes the stall.

## Separate finding — unrelated to the test failure

`RealBrowserManager.snapshot_recording_prefixes` is synchronous and does per-artifact `os.path.exists()` / `os.path.getsize()`:

```python
if not video_artifact.video_artifact_id or not path or not os.path.exists(path=path):
    return None
...
prefix_len=os.path.getsize(path)
```

It is called unawaited from the async cleanup path (`snapshots = app.BROWSER_MANAGER.snapshot_recording_prefixes(...)`). Blocking filesystem syscalls on an async path will stall the event loop **in production**, more so with many artifacts or network-backed storage. This **cannot** be the cause of the test failure — the test replaces `app` with a `MagicMock`, so it never touches a filesystem — but it seemed worth raising on its own merits.

## Ask

This is a repo-sync mirror PR, so a fix pushed to `repo-sync/main` gets reverted by the next upstream cut. Could you pick this up in the private repo? Either the ~1s loop block gets removed, or the bound gets adjusted if the suite's new weight makes 0.2s unrealistic.

No action has been taken on this PR: not closed, not merged, nothing pushed. Head is still `2460d04e`.

AronPerez:
Thanks @wintonzheng — no objection to dropping the wall-clock bound, and this isn't a request to revert. One measurement from our side that bears on what still guards this path.

**Final sample set on `2460d04e`: 6/6 failures** — 1.1056s, 0.9986s, 1.3598s, 1.3547s, 1.1174s, and your attempt 6 at 1.3875s. Control: `main` was 8/8 green on the same suite over the same window.

**The part worth knowing: `asyncio.wait_for(..., timeout=0.5)` never fired in any of the six.** The coroutine completed every time — only the monotonic delta exceeded the bound. So `wait_for` is not a backstop for this failure mode; it missed it 6/6. With the wall-clock assert removed, the ~1.0–1.39s of event-loop blocking is no longer detected by anything in this test.

To be fair to the change: `finalize.assert_not_awaited()` / `save.assert_not_awaited()` still cover the behavioural contract, and a genuine hang would still trip `wait_for`. Dropping a fragile wall-clock assertion from shared CI is reasonable on its own terms. The only reason I'm flagging it is that the stall was never explained, and it reproduces on this content while being absent on `main`.

Separately, still open from my previous comment: `RealBrowserManager.snapshot_recording_prefixes` does per-artifact `os.path.exists()` / `os.path.getsize()` synchronously and is called unawaited from the async cleanup path. That **cannot** be the test failure — the test replaces `app` with a `MagicMock` — but it will block the event loop in production, more so with many artifacts or network-backed storage.

Happy to leave it here; just wanted the decision to be an informed one.

AronPerez:
**Correction to my earlier comments — I got this wrong, and the causation claim should be withdrawn.**

I said the ~1s stall "reproduces on this content while being absent on `main`." That is now falsified.

`main` at `b37c56c` carries **the same `agent.py` content and the same wall-clock assertion** (the assert came back via #8456, which was cut upstream before #8449's removal landed) — and it **passes**: `25623 passed, 909 skipped, 1 xfailed in 740.07s`. That is *more* tests over a *longer* run than the PR head's `25233 passed` in 660–728s, which failed 6/6.

So the code is exonerated. The 6/6 measurement was real, but it was tied to the PR-run context, not to the diff.

**Where my reasoning broke:** "fails 6/6 on the PR head, `main` is 8/8 green" looked like a controlled comparison and wasn't. Those populations differed in at least three ways at once — the code, the workflow event (`pull_request` vs `push`, plausibly different runner pools), and the suite composition. I attributed the difference to the most visible variable. The only clean control was the same content in both contexts, and merging supplied it.

Two consequences worth stating:

- @wintonzheng's call to drop the wall-clock bound was **right**, and my earlier note arguing `wait_for` isn't a backstop, while factually true, was in service of a conclusion that doesn't hold. Apologies for the noise.
- My suite-weight/GC hypothesis is also dead — `main` runs 390 more tests and stays green.

**What still stands, independently of all the above:** `RealBrowserManager.snapshot_recording_prefixes` does per-artifact `os.path.exists()` / `os.path.getsize()` synchronously and is called unawaited from the async cleanup path. That was never the test failure — the test mocks `app` — and it isn't urgent. It's just a blocking-syscall-on-an-async-path observation, worth what it's worth.

Nothing here needs action. Correcting the record so nobody goes looking for a bug my earlier comment implied.

PR #8456, opened 2026-09-06
Let Copilot read and act on the page a credential run left instead of refusing on taint (SKY-15502)

andrewneilson:
Sync PR: 16712
Sync SHA: 84584bd3ba9d761ecd02e9f521216ae32c345f5a


andrewneilson:
⚠️ **Migration Alert**

The source PR included database migration changes in `alembic/versions/`.
Auto-generation was attempted but no migration was produced (tables not in this repo, or an error).

If these migrations affect open-source tables, please manually create corresponding migrations.

Sync PR: 16712

AronPerez:
@andrewneilson — heads-up on the CodeQL block here, with a suggested fix. **Not dismissing anything**; the call is yours.

**The alert is genuinely introduced by this PR.** `js/incomplete-url-substring-sanitization`, alert #373, above `main`'s current high baseline #363. `urlMayBeGoogleDrive` does not exist on `main`, so it isn't inherited.

**The pattern is real** — `workflowEditorUtils.ts:4335`:

```ts
candidate.toLowerCase().includes("drive.google.com")
```

matches `https://evil.com/?x=drive.google.com` and `https://drive.google.com.evil.com`.

**My read is that it isn't exploitable where it sits**, and I'd rather say so than imply a severity I don't think is there. It gates `showGoogleDriveSourceAccount` in `FileDownloadEditor.tsx:154` — whether a form field is *displayed* in the editor. It isn't an access-control, trust, or credential-routing decision; `googleCredentialId` is a separate OR'd condition. The function is an explicit heuristic that already returns true for any string containing `{{` and for any bare alphanumeric id, so it's deliberately over-inclusive and was never a sanitizer. Worst case is the account selector showing when it arguably shouldn't.

**That said, fixing looks cheaper than dismissing** — a hostname comparison clears the alert outright and keeps the existing behaviour, including scheme-less input:

```ts
function urlMayBeGoogleDrive(url: string): boolean {
  const candidate = url.trim();
  if (candidate.includes("{{") || bareDriveFileId.test(candidate)) {
    return true;
  }
  try {
    const host = new URL(
      /^[a-z][a-z0-9+.-]*:\/\//i.test(candidate) ? candidate : `https://${candidate}`,
    ).hostname.toLowerCase();
    return host === "drive.google.com" || host.endsWith(".drive.google.com");
  } catch {
    return false;
  }
}
```

Untested against your suite — treat it as a sketch of the shape rather than a patch.

Since this is a repo-sync mirror PR, anything applied to `repo-sync/main` gets reverted by the next upstream cut, so it needs to land in the private repo.

No action taken here: not merged, not closed, no rerun, no dismissal, nothing pushed. Held pending your decision.

PR #8459, opened 2026-09-06
Recognize JS-wired submit controls in the explicit-submit click bypass (SKY-15195)

LawyZheng:
Sync PR: 16715
Sync SHA: b5ec480acdb581f8822dfba3756a1c4ccb838568

### DIFF
--- a/skyvern-frontend/src/routes/workflows/editor/panels/BlockConfigForm/FileDownloadBlockForm.test.tsx
+++ b/skyvern-frontend/src/routes/workflows/editor/panels/BlockConfigForm/FileDownloadBlockForm.test.tsx
@@ -511,18 +511,6 @@ describe("FileDownloadBlockForm (SKY-9361)", () => {
     ).toBeDefined();
   });
 
-  test("does not treat a URL containing the Drive hostname as a Drive link", () => {
-    setFileDownloadNode("d1", {
-      downloadTarget: "sftp",
-      url: "https://example.com/drive.google.com/file/d/abc123/view",
-    });
-    render(<FileDownloadBlockForm blockId="d1" />);
-
-    expect(
-      screen.queryByText("Google Drive Source Account (Optional)"),
-    ).toBeNull();
-  });
-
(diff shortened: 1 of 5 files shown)


AronPerez:
@wintonzheng @andrewneilson — this is the **third** sync PR in a row to revert the Google Drive host validation from #8456, and it will keep happening until the fix lands upstream.

**The loop:** #8456's fix merged to OSS `main` at 01:14:05Z. Every mirror cut from the private repo *after* that but *before* the fix exists upstream carries the pre-fix `urlMayBeGoogleDrive` and deletes `FileDownloadBlockForm.test.tsx`'s regression test.

| PR | Status |
|---|---|
| #8458 | reverted it — patched, merged clean |
| #8459 | reverting it — patched, this PR |
| next cut | will revert it again |

**The durable fix is one-way:** land the hostname check in the **upstream private repo**. Until then each mirror needs a manual restore, and the window between a green CodeQL and a merge is where the vulnerable form can slip back onto `main`.

Worth noting on that last point: on #8458 CodeQL returned **neutral** on a head that provably carried the vulnerable substring form. Here it correctly failed with #373. So a green or neutral CodeQL is not sufficient evidence the reversion is absent — the file contents are.

The fix to port upstream is exactly what's on `main` now:

```ts
function urlMayBeGoogleDrive(url: string): boolean {
  const candidate = url.trim();
  if (candidate.includes("{{") || bareDriveFileId.test(candidate)) {
    return true;
  }
  try {
    const parsed = new URL(candidate);
    return (
      parsed.protocol === "https:" &&
      parsed.hostname === "drive.google.com" &&
      parsed.pathname.startsWith("/file/d/")
    );
  } catch {
    return false;
  }
}
```

plus the `FileDownloadBlockForm.test.tsx` case that covers it.

**On this PR specifically:** its actual contribution (SKY-15195, the JS-wired submit-control changes in `agent_functions.py` / `handler.py` and the new 112-line test) is untouched and fine. Only the two frontend files are being restored to `main`'s version. It's briefly in draft as a merge brake and will be marked ready once the restore is verified.

PR #8618, opened 2026-09-19
Register per-org OpenRouter gpt-5.6-luna configs and fix the responses-bridge guard they expose (SKY-16623)

pedrohsdb:
Sync PR: 17245
Sync SHA: 0208d8d8a0ba76ea6cfaea4530835061b702d79b


wintonzheng:
CodeQL failed here because the sync re-copied `tests/unit/test_forge_log_redaction.py` and `skyvern/forge/sdk/services/twilio_service.py` from cloud main, which does not carry Skyvern-AI/skyvern-cloud#17250 yet. Re-applied both fixes in cde7cfc (same content as that PR). This will recur on each new rolling PR until #17250 merges.

PR #8624, opened 2026-09-21
Rename CodeBlock redaction test constant so OSS CodeQL stops flagging it (SKY-16595)

andrewneilson:
Sync PR: 17315
Sync SHA: af317fd4f751bbcf7c6221d2c209cee75e2234b2


andrewneilson:

Sync PR: 17315

